PR #8, opened 2024-06-14
Ci revert workarounds

LeonMatthesKDAB:
Now that Windows CI Runners seem to be back to normal, test whether we can revert the workarounds.

See also: https://github.com/gabime/spdlog/issues/3107


narnaud (approved):
Yes! thank you

Issue #3, opened 2024-06-06
CI ToDo List


LeonMatthesKDAB:
As mentioned in #1, apart from setting up the initial CI for testing, there's a few more ToDos:

- [x] enable cached builds
- [x] Add tests for mfc-utils and photonwidgets
- [x] Clazy check
- [x] pre-commit checks (including clang-format)
- [ ] clang-tidy check

Additional automations:
- [x] Release-please
- [x] Automatic documentation deployment?

LeonMatthesKDAB:
This is mostly done now, and we don't want to enable clang-tidy at the moment, so let's close this issue :)

PR #9, opened 2024-06-14
ci: Automate updates to documentation

LeonMatthesKDAB:
Part of #3

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -32,9 +32,7 @@ jobs:
         # Delete the ref to avoid keeping history.
         git update-ref -d refs/heads/gh-pages
         rm -rf *
-        # We want the documentation in a subfolder, so we can add a landing page or similar
-        mkdir -p docs/latest/
-        mv ../site/* docs/latest/
+        mv ../site/* .
         git add .
         git commit -m "Deploy $GITHUB_SHA to gh-pages"
         git push --force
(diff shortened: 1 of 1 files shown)


.github/workflows/documentation.yml
```diff
@@ -0,0 +1,40 @@
...
+    branches:
+      - main
+    paths:
+      - docs/**
+      - mkdocs.yml
+
+jobs:
+  mkdocs:
+    # Ensure that the tag was from the main branch and not a backport
+    if: github.event.base_ref == 'refs/heads/main'
```
narnaud, line 15: I wonder if we could do that only if a file in docs has been changed, that would be perfect.
Feel free to mark as done if not possible.

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,40 @@
...
+      run: |
+        mkdocs build
+        git worktree add gh-pages gh-pages
+        git config user.name "Deploy from CI"
+        git config user.email ""
+        cd gh-pages
+        # Delete the ref to avoid keeping history.
+        git update-ref -d refs/heads/gh-pages
+        rm -rf *
+        # We want the documentation in a subfolder, so we can add a landing page or similar
```
narnaud, line 35: Not really, there's already a landing page on the documentation.
Let's not overcomplexify things, specially as it will look wrong compare to the rest. Just put it at the root of the website.

.github/workflows/documentation.yml
LeonMatthesKDAB: That should already be the case (see the `on>push>paths` key, where I've specified the paths to watch for)

.github/workflows/documentation.yml
LeonMatthesKDAB: Done

narnaud (approved):

PR #13, opened 2024-06-17
Refactor script dialog for pregression

narnaud:
- Refactor the ScriptDialogItem progression API, to ahve something simpler to use
- Add 2 examples to show how to do that in scripts
- Remove the uiFilePath property from ScriptDialog


narnaud (commented):
I have a patch to fix mfc-utils too. I will add to this review if we agree on the API, for now the test won't pass.

Montel (approved):

PR #14, opened 2024-06-17
fix: message map extraction when namespaces are used

narnaud:
The message map may be in a namespace, we can't assume it is top-level.


Montel:
Is it possible to add a autotest ?

narnaud:
> Is it possible to add a autotest ?

You are correct, I should add autotest for fixes, will do.

LeonMatthesKDAB:
@narnaud Have you checked how this affects performance? I think I remember trying something similar and it caused the whole query to be significantly slower, because it's now giving a lot more possible results that need to be filtered by the predicates?
But maybe that was on a different query.

narnaud:
> @narnaud Have you checked how this affects performance? I think I remember trying something similar and it caused the whole query to be significantly slower, because it's now giving a lot more possible results that need to be filtered by the predicates? But maybe that was on a different query.

I think the problem was when we were using it for contextual scripting... performances are less an issue when used in script.
But no, I didn't test it for performances.

Maybe I could add an optional test for the namespace, not sure how to do that (will check later).

narnaud:
All fixed:
- the query test for MESSAGE_MAP either at top-level or in a namespace
- added a test to ensure it's working

Montel (approved):

Issue #37, opened 2024-06-19
Add support for QML TreeSitter


narnaud:
We need to integrate the QML grammar into TreeSitter, most likely using a 3rd party.
I expect that once done, we should have access to the different symbols into the QML file.

PR #46, opened 2024-06-21
core: Add support for QmlJs via tree-sitter-qmljs

itzurabhi:
Add QML TreeSitter support in build.

Part of #37


narnaud (commented):
Looks good, just waiting for @LeonMatthesKDAB approval.
I'm surprised it's so easy, nice work ;)

LeonMatthesKDAB (approved):
Some context on this PR:
@itzurabhi and I discussed for a while which tree-sitter grammar to use for QML.

The one from yuja stood out as the most complete, it is however still not "official" in any way shape or form.
But because there is no official Tree-sitter grammar, we have to rely on community bindings like this one.

The main reasons we chose this grammar:

### Testing

The author tests their grammar against the examples in the Qt-declarative repository, which Qt also uses to test their own parser.
There are a few [known failures](https://github.com/yuja/tree-sitter-qmljs/blob/master/examples/known-failures.txt), but it's not that many.

### Inclusion of Typescript grammar

QML is made up of half-Javascript, half custom syntax.
This grammar uses the typescript grammar for the javascript part, on which we should be able to rely on going forward, as Typescript is of course very popular.
So in the worst case, we may have to find the right names for the QML part of the our tree-sitter queries, but can keep any conversions that we do in JS code, as they just rely on the typescript grammar.

So all in all, with these caveats in mind, I believe this is a good way forward on which we should be able to rely until (if at all) Qt provide their own tree-sitter grammar.

Also note that this PR doesn't yet add support for any QML documents, but that's a reasonably easy step to achieve later.

PR #71, opened 2024-06-26
feat: introduce Qml to parser

itzurabhi:
add the Qml tree sitter language support to all places except test files. 

part of #37

### DIFF
--- a/src/core/qmldocument.cpp
+++ b/src/core/qmldocument.cpp
@@ -13,8 +13,9 @@
 namespace Core {
 
 QmlDocument::QmlDocument(QObject *parent)
-    : TextDocument(Type::Qml, parent)
+    : CodeDocument(Type::Qml, parent)
 {
 }
 
+QmlDocument::~QmlDocument() = default;
 }
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -622,7 +622,7 @@ void MainWindow::updateActions()
     const bool lspEnabled = codeDocument && codeDocument->hasLspClient();
     ui->actionFollowSymbol->setEnabled(lspEnabled);
     ui->actionSwitchDeclDef->setEnabled(lspEnabled);
-    ui->actionTreeSitterInspector->setEnabled(lspEnabled);
+    ui->actionTreeSitterInspector->setEnabled(codeDocument != nullptr);
 
     const bool cppEnabled = codeDocument && qobject_cast<Core::CppDocument *>(document);
     ui->actionSwitchHeaderSource->setEnabled(cppEnabled);
--- a/src/treesitter/parser.cpp
+++ b/src/treesitter/parser.cpp
@@ -10,6 +10,7 @@
 
 #include "parser.h"
 #include "tree.h"
+#include "treesitter/languages.h"
 
 #include <tree_sitter/api.h>
 #include <utility>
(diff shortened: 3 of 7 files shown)


CMakeLists.txt
```diff
@@ -135,6 +135,14 @@ add_custom_command(
           ${CMAKE_BINARY_DIR}/compile_commands.json compile_commands.json
   WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}/test_data/projects/cpp-project)
 
+  add_custom_command(
```
itzurabhi, line 138: will remove this, part of work in progress.

test_data/projects/CMakeLists.txt
```diff
@@ -8,3 +8,4 @@
 #
 
 add_subdirectory(cpp-project)
+add_subdirectory(qml-project)
```
itzurabhi, line 11: will remove this

src/gui/mainwindow.cpp
```diff
@@ -619,10 +619,11 @@ void MainWindow::updateActions()
     ui->actionExecuteAPI->setEnabled(textDocument != nullptr);
 
     auto *codeDocument = qobject_cast<Core::CodeDocument *>(document);
+    const bool qmlEnabled = codeDocument && qobject_cast<Core::QmlDocument *>(document);
     const bool lspEnabled = codeDocument && codeDocument->hasLspClient();
     ui->actionFollowSymbol->setEnabled(lspEnabled);
     ui->actionSwitchDeclDef->setEnabled(lspEnabled);
-    ui->actionTreeSitterInspector->setEnabled(lspEnabled);
+    ui->actionTreeSitterInspector->setEnabled(lspEnabled || qmlEnabled);
```
narnaud, line 626: checking that it's a codedocument should be enough, as it will have a TreeSitter backend
No need for `qmlEnabled`

src/core/qmldocument.cpp
```diff
@@ -9,12 +9,14 @@
 */
 
 #include "qmldocument.h"
+#include "astnode.h"
```
narnaud, line 12: Why this change?

src/treesitter/parser.h
```diff
@@ -38,6 +39,8 @@ class Parser
 
     const TSLanguage *language() const;
 
+    static TSLanguage *getTSLanguageForDocumentType(Core::Document::Type type);
```
narnaud, line 42: as the Parser is using `language` as a name, we could reduce the static name to `getLanguage(Core::Document::Type type)`, no need for TS or re-adding the parameter type.

src/treesitter/parser.cpp
```diff
@@ -62,4 +63,18 @@ const TSLanguage *Parser::language() const
     return ts_parser_language(m_parser);
 }
 
+TSLanguage *Parser::getTSLanguageForDocumentType(Core::Document::Type type)
+{
+    switch (type) {
+    case Core::Document::Type::Qml:
+        return tree_sitter_qmljs();
+    // FIXME: make sane default for other documents.
```
narnaud, line 71: In fact, I would put the Q_UNREACHABLE under `default:`, this should be a good default.
Because as it is, it should be impossible to reach that for now.

narnaud (commented):
It's getting there, quite happy to see the inspector working with a QML file.

src/core/qmldocument.cpp
itzurabhi: editor doing autoimports :(

src/treesitter/parser.h
itzurabhi: done

narnaud (approved):
Thank you!

PR #79, opened 2024-07-01
feat: Add symbol listing for Qml

itzurabhi:
symbol listing for Qml added.


src/core/codedocument_p.h
```diff
@@ -25,6 +26,25 @@ class CodeDocument;
 
 class TreeSitterHelper
 {
+public:
+    using SymbolQueryFunction = std::function<QList<Core::Symbol *>(CodeDocument *const)>;
```
LeonMatthesKDAB, line 30: Minor nitpick: But maybe this should use `const CodeDocument &` as argument, as we don't expect the CodeDocument to be `nullptr`.

src/core/codedocument_p.h
```diff
@@ -25,6 +26,25 @@ class CodeDocument;
...
+
+    SymbolQueryFunction functionSymbols = [](CodeDocument *const) -> QList<Core::Symbol *> {
```
LeonMatthesKDAB, line 32: Any reason why we use 4 different functions instead of a single `querySymbols` function here?

That querySymbols function can internally still use as many functions as is sensible, but that would make it less akward that we have to assign the UIObjects to classSymbols in the QML document.
```
helper()->classSymbols = QmlDocument::queryUIObjectSymbols;
```

src/core/qmldocument.cpp
```diff
@@ -9,13 +9,69 @@
...
+        if (!match.get("return").isValid()) {
+            // No return type, this is a Constructor/Destructor
+            // Clangd also assigned the Constructor kind to Destructors, so we'll do the same
+            kind = Symbol::Kind::Constructor;
+        } else if (match.get("name").text().contains("::")) {
+            // This is a bit of a guesstimate, but if the function name contains "::", it's likely a method.
+            // It may also be a member of a namespace, but this information isn't really available unless we try
+            // to resolve the original declaration.
+            kind = Symbol::Kind::Method;
+        }
```
LeonMatthesKDAB, line 56: I don't think this makes sense for a QML document?

src/core/codedocument_p.h
```diff
@@ -25,6 +26,25 @@ class CodeDocument;
...
+        return {};
+    };
+
+    SymbolQueryFunction memberSymbols = [](CodeDocument *const) -> QList<Core::Symbol *> {
+        return {};
+    };
+
+    SymbolQueryFunction enumSymbols = [](CodeDocument *const) -> QList<Core::Symbol *> {
+        return {};
+    };
```
narnaud, line 46: To reduce clutter here, maybe you can initialize them in the constructor.
Or create a macro:
```
    SymbolQueryFunction functionSymbols = EMPTY_SYMBOL_LIST;
    SymbolQueryFunction classSymbols = EMPTY_SYMBOL_LIST;
    SymbolQueryFunction memberSymbols = EMPTY_SYMBOL_LIST;
    SymbolQueryFunction enumSymbols = EMPTY_SYMBOL_LIST;
```
Slightly easier to read

src/core/codedocument_p.h
```diff
@@ -25,6 +26,25 @@ class CodeDocument;
...
+
 public:
```
narnaud, line 48: no need for this public here

src/core/cppdocument.h
```diff
@@ -111,6 +112,11 @@ public slots:
                                const QString &newClassBaseName);
     void changeBaseClassForwardInclude(const QString &originalClassBaseName, const QString &newClassBaseName);
 
+    static auto queryFunctionSymbols(CodeDocument *const) -> QList<Core::Symbol *>;
+    static auto queryClassSymbols(CodeDocument *const) -> QList<Core::Symbol *>;
+    static auto queryMemberSymbols(CodeDocument *const) -> QList<Core::Symbol *>;
+    static auto queryEnumSymbols(CodeDocument *const) -> QList<Core::Symbol *>;
```
narnaud, line 118: Let's not use auto here, doesn't add much.
Also, static private doesn't make much sense, move it all to the C++ file, in an anonymous namespace.

src/core/qmldocument.h
```diff
@@ -21,6 +22,11 @@ class QmlDocument : public CodeDocument
 public:
     explicit QmlDocument(QObject *parent = nullptr);
     ~QmlDocument() override;
+
+private:
+    static auto queryUIObjectSymbols(CodeDocument *const) -> QList<Core::Symbol *>;
+    static auto queryFunctionSymbols(CodeDocument *const) -> QList<Core::Symbol *>;
+    static auto queryMemberSymbols(CodeDocument *const) -> QList<Core::Symbol *>;
```
narnaud, line 29: Same here, move it all to the C++ file, in an anonymous namespace.

narnaud (commented):
This is on the right track. I didn't test it (I wonder if you see the symbols in the GUI yet), but the architecture definitely makes sense.

src/core/qmldocument.cpp
```diff
@@ -9,13 +9,69 @@
...
+
+    return kdalgorithms::transformed<QList<Symbol *>>(functions, function_to_symbol);
+}
+
+auto QmlDocument::queryMemberSymbols(CodeDocument *const document) -> QList<Core::Symbol *>
+{
+    auto members = document->query(QString(R"EOF(
+        (ui_binding
+        name: (_) @name @selectionRange
+        value: (expression_statement(_)@type)
```
LeonMatthesKDAB, line 68: Note that this won't find members that assign another component.

e.g. in an ApplicationWindow it won't find the `footer`.

```qml
ApplicationWindow {
    footer: RowLayout {
        // ...
    }
}
```
As that is a `ui_binding`, not an `expression_statement`.

src/core/qmldocument.cpp
```diff
@@ -9,13 +9,69 @@
...
+                (ui_object_definition type_name : (_) @name @selectionRange) @range
+            )EOF"));
+    auto objectsToSymbol = [document](const QueryMatch &match) {
+        return Symbol::makeSymbol(document, match, Symbol::Kind::Class);
+    };
+    return kdalgorithms::transformed<QList<Symbol *>>(uiObjectDefinitions, objectsToSymbol);
+}
+auto QmlDocument::queryFunctionSymbols(CodeDocument *const document) -> QList<Core::Symbol *>
+{
+    auto functions = document->query(QString(R"EOF((function_declaration
```
LeonMatthesKDAB, line 39: These kinds of S-expressions are very similar to LISP code, so should be formatted and read by indentation (see also: https://ashok-khanna.medium.com/formatting-lisp-5e28020b8bac )

Please format this in a lisp-like style
e.g. I would format it like this:
```lisp
(function_declaration
  name: (_) @name @selectionRange
  parameters: (_) @parameters
  body: (_) @body) @range
```
with `R"EOF(` on a separate line

Same with the ui_binding query in queryMemberSymbols

LeonMatthesKDAB (commented):
Overall looks like a reasonable way forward :+1: 

One additional thing to consider though is that the FunctionSymbol class also runs a few small queries to figure out the names and types of parameters (see the argumentsFromQueryMatch function).

However, that is of course also C++-specific and won't work with QML.
We should try to find a solution for injecting the right behavior there as well. 

Also added some minor nitpicks.

src/core/codedocument_p.h
itzurabhi: `Symbol::makeSymbol` takes `parent` as a `QObject*` should I change that as well ?

itzurabhi:
> This is on the right track. I didn't test it (I wonder if you see the symbols in the GUI yet), but the architecture definitely makes sense.

yes we can

src/core/codedocument_p.h
narnaud: No, that's ok I think. It's jsut here we want to make sure the CodeDocument is not null

src/core/qmldocument.cpp
```diff
@@ -9,12 +9,76 @@
...
+    auto functions = document->query(queryString);
+
+    auto function_to_symbol = [document](const QueryMatch &match) {
+        return Symbol::makeSymbol(document, match, Symbol::Kind::Function);
+    };
+
+    return kdalgorithms::transformed<QList<Symbol *>>(functions, function_to_symbol);
+}
+
+auto queryMemberSymbols(CodeDocument *const document) -> QList<Core::Symbol *>
```
narnaud, line 47: s/Member/Property, if it is what I think it is.

src/core/qmldocument.cpp
```diff
@@ -9,12 +9,76 @@
...
+        return Symbol::makeSymbol(document, match, Symbol::Kind::Object);
+    };
+    return kdalgorithms::transformed<QList<Symbol *>>(uiObjectDefinitions, objectsToSymbol);
+}
+auto queryFunctionSymbols(CodeDocument *const document) -> QList<Core::Symbol *>
+{
+    const auto queryString = QString(R"EOF(
+        (function_declaration
+        name:(_) @name @selectionRange
+        parameters: (_) @parameters
```
LeonMatthesKDAB, line 34: Ideally this should capture each individual parameter as `@parameter`, so that The FunctionSymbol can then extract them.

LeonMatthesKDAB (commented):
The separation of concerns looks clean :+1: 

What I'd like to see addressed before merging:
1. Find a way to make `FunctionSymbol::arguments` compatible with Qml.
    * Currently it's not capturing the individual parameters as `@parameter`, so the FunctionSymbol will always report 0 arguments
    * The types of the arguments would be extracted incorrectly in the `FunctionSymbol::argumentsFromQueryMatch` function, which would have to also differentiate between languages.
2. Add tests for all types of QML symbol extraction. Currently this is entirely untested.

I guess we could also do these in follow-up PRs, but I don't expect much conflict in this part of the codebase at the moment, so it would be preferable to have everything in before merging.

LeonMatthesKDAB:
Hm, I also took another look at the ClassSymbol. I guess that also won't work correctly, as it will probably associate all members from all sub-objects into itself.
It currently doesn't check whether there's another classsymbol that encompasses sub-members :thinking:

tests/tst_qmldocument.cpp
```diff
@@ -0,0 +1,52 @@
...
+    Q_OBJECT
+
+private slots:
+    void initTestCase() { Q_INIT_RESOURCE(core); }
+    void testAllSymbols()
+    {
+        Test::testQmlDocument("", "tst_symbols.qml", [](auto *document) {
+            auto allSymbols = document->symbols();
+
+            {
```
LeonMatthesKDAB, line 20: What's the point of the curly brackets here and around the other sections?

They make the test very long, why not just separate the sections with one line of whitespace?

test_data/tst_symbols.qml
```diff
@@ -0,0 +1,49 @@
...
+
+TestCase {
+    name: "Project"
+    id : sane
+    age : 12
+
+    sub : Another {
+        id : 123
+    }
+
```
LeonMatthesKDAB, line 23: Let's also add some child components that aren't assigned to properties.
That's pretty much the use case for Qml, but isn't included in the test.

E.g. just a random `Item` or a Layout with some `Rectangles` would do. The QML doesn't actually have to make any sense.
But we want to exercise our queries as much as possible.

test_data/tst_symbols.qml
```diff
@@ -0,0 +1,49 @@
...
+        var files = Project.allFiles();
+        compare(files.length, 10)
+        compare(files[0], "MFC_UpdateGUI.rc")
+
+        var rcFiles = Project.allFilesWithExtension("rc");
+        compare(rcFiles.length, 1)
+        compare(rcFiles[0], "MFC_UpdateGUI.rc")
+    }
+
+    function test_open() {
```
LeonMatthesKDAB, line 36: Maybe add some type arguments to some of the methods and add type annotations to some of them.

You can probably also reduce the code within the methods to something shorter, as it's not really relevant for testing.

This test file should try to exercise as many edge cases as possible.
It's about including as many kinds of symbols as you can think of.
The QML itself doesn't have to do anything useful.

tests/tst_qmldocument.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "core/utils.h"
+#include <algorithm>
+
+class TestQmlDocument : public QObject
+{
+    Q_OBJECT
+
+private slots:
+    void initTestCase() { Q_INIT_RESOURCE(core); }
+    void testAllSymbols()
```
LeonMatthesKDAB, line 15: Overall I would like to see more tests on the symbol contents themselves.
This test just counts how many symbols there are, which is good to have.
But we should have at least a few tests that actually check whether what is in the symbols is even correct.

Maybe take a look at how you can provide data to a test, like we do in tst_symbol.cpp, so you can test a bunch of symbols for their contents without too much code duplication.

LeonMatthesKDAB (commented):
Going in the right direction, but the testing should be more thorough.

LeonMatthesKDAB (approved):
Looks good. We'll want to follow up on this with fixes for FunctionSymbol and ClassSymbol, but a very good start :partying_face:

Issue #54, opened 2024-06-22
Review public API


narnaud:
For each API, ensure there is:
- documentation (property or method)
- ensure the documentation is properly rendered
- proper `LOG`

List of API to review:
- [x] Document
- [x] Project
- [x] CodeDocument
    - [x] CodeDocument
    - [x] ClassSymbol
    - [x] FunctionArgument
    - [x] FunctionSymbol
    - [x] QueryCapture
    - [x] QueryMatch
    - [x] Symbol
    - [x] AstNode
- [x] CppDocument
    - [x] CppDocument
    - [x] DataExchange
    - [x] DataExchangeEntry
    - [x] MessageMap
    - [x] MessageMapEntry
- [x] QtTsDocument
    - [x] QtTsDocument
    - [x] QtTsMessage
- [x] QtUiDocument
    - [x] QtUiDocument
    - [x] QtUiWidget
- [x] RcDocument
    - [x] RcDocument
    - [x] Action
    - [x] Asset
    - [x] Menu
    - [x] MenuItem
    - [x] Ribbon
    - [x] RibbonCategory
    - [x] RibbonContext
    - [x] RibbonElement
    - [x] RibbonMenu
    - [x] RibbonPanel
    - [x] Shortcut
    - [x] String
    - [x] ToolBar
    - [x] ToolBarItem
    - [x] Widget
- [x] TextDocument
    - [x] TextDocument
    - [x] Mark
    - [x] RangeMark
    - [x] TextLocation
    - [x] TextRange
- [x] Items
    - [x] Script
    - [x] ScriptDialog
- [x] Utilities
    - [x] Dir
    - [x] File
    - [x] FileInfo
    - [x] Message
    - [x] Settings
    - [x] UserDialog
    - [x] Utils
    - [x] QDirValueType
    - [x] QFileInfoValueType

narnaud:
All done

PR #80, opened 2024-07-01
Review some API and update documentation

narnaud:
- docs: Review the `Document` and `Project` API
- docs: Review the item's API
- fix(docs): Fix an issue with `array<>` in doc
- docs: remove old LSP API
- docs: review the Utils API

Related to #54

### DIFF
--- a/src/core/codedocument.cpp
+++ b/src/core/codedocument.cpp
@@ -206,7 +206,7 @@ QString CodeDocument::hover(int position, std::function<void(const QString &)> a
 std::pair<QString, std::optional<TextRange>> CodeDocument::hoverWithRange(
     int position, std::function<void(const QString &, std::optional<TextRange>)> asyncCallback /*  = {} */) const
 {
-    LOG("CodeDocument::hover");
+    spdlog::debug("CodeDocument::hover");
 
     if (!checkClient())
         return {"", {}};
@@ -263,7 +263,7 @@ std::pair<QString, std::optional<TextRange>> CodeDocument::hoverWithRange(
 
 Core::TextLocationList CodeDocument::references(int position) const
 {
-    LOG("CodeDocument::references");
+    spdlog::debug("CodeDocument::references");
 
     if (!checkClient()) {
         return {};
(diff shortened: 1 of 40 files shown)


src/core/userdialog.cpp
```diff
@@ -50,6 +51,7 @@ UserDialog::UserDialog(QQmlEngine *parent)
  */
 QJSValue UserDialog::getOpenFileName(const QString &caption, const QString &dir, const QString &filters)
 {
+    LOG("UserDialog::getOpenFileName", caption, dir, filters);
```
LeonMatthesKDAB, line 54: Should this use LOG_RETURN as well?
Same for the other UserDialog functions :thinking:

src/core/codedocument.cpp
```diff
@@ -205,8 +205,6 @@ QString CodeDocument::hover(int position, std::function<void(const QString &)> a
 std::pair<QString, std::optional<TextRange>> CodeDocument::hoverWithRange(
     int position, std::function<void(const QString &, std::optional<TextRange>)> asyncCallback /*  = {} */) const
 {
-    LOG("CodeDocument::hover");
```
LeonMatthesKDAB, line 208: I would prefer to replace this with `spdlog::debug()`.

If there are any issues with this code, we can easier spot the root cause instead of having to debug random LSP failures that occur apparently out of nowhere :)

Same with the other LSP methods.

LeonMatthesKDAB (changes requested):
I'd prefer to not completely get rid of the logging, for the LSP functions, plus another minor nitpick, but otherwise looks good :)

src/core/userdialog.cpp
narnaud: Not really, the LOG_RETURN is mostly used for API that is used inside the GUI, to be able to create a script from the history log.
For example, when doing a find on the current selected text, teh LOG_RETURN is used to match the return value of `TextDocument::selectedText` and the parameter of `TextDocument::find`.

Here, those methods are only used in scripts, and never called from teh GUI, so we don't need that. We still need the log to see what's going on when running the script.

src/core/codedocument.cpp
narnaud: Good idea, fixed.

src/core/userdialog.cpp
LeonMatthesKDAB: Hm, okay. But then shouldn't we just use `spdlog::trace` or `spdlog::debug`?

narnaud: Ok, you want to know the return value returned. LOG_RETURN won't help here, as it's logging the "tag" of the return value, not the return value itself.

I've added spdlog::debug, this way we will see it even if we remove the trace (that's where the `LOG` are logged).

Just review the last commit, rest has not changed.

LeonMatthesKDAB (approved):

PR #82, opened 2024-07-02
fix: modification of the setTranslation function for the case where the node is untranslated

smnperret-kdab:
refer to PR #87 for upcoming commits

### DIFF
--- a/src/core/qttsdocument.cpp
+++ b/src/core/qttsdocument.cpp
@@ -297,10 +297,11 @@ QString QtTsMessage::translation() const
     return QString::fromUtf8(m_message.child("translation").text().as_string());
 }
 
-void QtTsMessage::setTranslation(const QString &translate)
+void QtTsMessage::setTranslation(const QString &translation)
 {
-    LOG("QtTsMessage::setTranslation", translate);
-    m_message.child("translation").set_value(translate.toUtf8().constData());
+    LOG("QtTsMessage::setTranslation", translation);
+    m_message.remove_child("translation");
+    m_message.child("translation").text().set(translation.toUtf8().constData());
     qobject_cast<QtTsDocument *>(parent())->setHasChanged(true);
     Q_EMIT translationChanged();
 }
--- a/tests/tst_qttsdocument.cpp
+++ b/tests/tst_qttsdocument.cpp
@@ -126,6 +126,19 @@ private slots:
         }
     }
 
+    void changeTranslation()
+    {
+        Core::QtTsDocument document;
+        document.load(Test::testDataPath() + QStringLiteral("/tst_qttsdocument/language_translation.ts"));
+        const auto messages = document.messages();
+        for (const auto &message : messages) {
+            message->setTranslation("new translation");
+        }
+        for (const auto &message : messages) {
+            QCOMPARE(message->translation(), "new translation");
+        }
+    }
+
     void createFromEmptyFile()
     {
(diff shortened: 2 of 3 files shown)


src/core/qttsdocument.cpp
```diff
@@ -297,10 +297,12 @@ QString QtTsMessage::translation() const
...
 
-void QtTsMessage::setTranslation(const QString &translate)
+void QtTsMessage::setTranslation(const QString &translation)
 {
-    LOG("QtTsMessage::setTranslation", translate);
-    m_message.child("translation").set_value(translate.toUtf8().constData());
+    LOG("QtTsMessage::setTranslation", translation);
+    m_message.remove_child("translation");
+    pugi::xml_node translationChild = m_message.append_child("translation"); // add translation
+    translationChild.append_child(pugi::node_pcdata).set_value(translation.toUtf8().constData());
```
narnaud, line 305: I found the right way:
```
    m_message.child("translation").text().set(translation.toUtf8().constData());
```
No need to remove and re-add a node...

tests/tst_qttsdocument.cpp
```diff
@@ -126,6 +126,18 @@ private slots:
         }
     }
 
+    void changeTranslation()
+    {
+        Core::QtTsDocument document;
+        document.load(Test::testDataPath() + QStringLiteral("/tst_qttsdocument/language_translation.ts"));
+        for (const auto &message : document.messages()) {
```
narnaud, line 133: `std::as_const(document.messages())` to cancel clazy warning

tests/tst_qttsdocument.cpp
```diff
@@ -126,6 +126,18 @@ private slots:
...
+            message->setTranslation("new translation");
+        }
+        for (const auto &message : document.messages()) {
```
narnaud, line 136: Here too: `std::as_const(document.messages())` to cancel clazy warning

src/core/qttsdocument.cpp
dfaure-kdab: You'd have to remove the type attribute too (type="unfinished"). Not sure if that's simpler in the end. But text() certainly looks good.

tests/tst_qttsdocument.cpp
dfaure-kdab: Can't use as_const on an rvalue. This needs a const local variable.

src/core/qttsdocument.cpp
```diff
@@ -297,10 +297,11 @@ QString QtTsMessage::translation() const
     return QString::fromUtf8(m_message.child("translation").text().as_string());
 }
 
-void QtTsMessage::setTranslation(const QString &translate)
+void QtTsMessage::setTranslation(const QString &translation)
 {
-    LOG("QtTsMessage::setTranslation", translate);
-    m_message.child("translation").set_value(translate.toUtf8().constData());
+    LOG("QtTsMessage::setTranslation", translation);
+    m_message.remove_child("translation");
```
narnaud, line 303: I was unclear, please remove this line, only keep the line below.
Also add: `m_message.child("translation").remove_attribute("type");`

The tests will magically work if you do that. The reason they are failing is that you are removing the translation node, but you are not adding a new one: `m_message.child("translation")` returns an empty node and all the operations after that are NOOP.

smnperret-kdab:
refer to PR #87 for upcoming commits

PR #84, opened 2024-07-02
refactor: reduce LSP dependencies in Core

### LINKED ISSUE #67: Refactor the text range/mark API
This is related to #54 , we currently have 4 different classes to handle position in text:
- simple `int`: that's the general position in a `TextDocument`
- `TextRange`: simple classes with start/end position
- `TextLocation`: contains a `TextRange` + a document
    - use only for `CodeDocument::references`
    - shouldn't be documented as it's not even public API
- `Mark`: position in a document, which is updated based on edits
- `RangeMark`: as the name suggest, that's a range of 2 marks, for start/end
    - very similar of `TextRange` and `TextLocation`, except for the auto update

I would like to reduce to 3 different classes:
- simple `int`: we keep this one for recording script, it's perfect
- `Cursor`: position in a document (ex `Mark`)
- `CursorRange`: range of cursor (ex `RangeMark`)

Cursor is a better term and more explicit, and I don't think we need the non-document aware version:
- either you want it for reading something, and it's not a problem
- or you want to keep it longer after edit, meaning you want updates.

narnaud:
- remove lsp_utils in knut-lsp
- create a lsp_utils with lsp methods in Core::Utils namespace
- cleanup methods from TextLocation and CodeDocument

Fix #67


src/core/lsp_utils.h
```diff
@@ -0,0 +1,38 @@
...
+#include <QString>
+#include <string>
+
+namespace Core {
+class TextDocument;
+}
+
+namespace Core::Utils {
+
+std::string documentToUri(const TextDocument *textDocument);
```
LeonMatthesKDAB, line 26: The implementation expects the textDocument to not be a nullptr. So prefer to use a reference instead of a pointer.

src/core/CMakeLists.txt
```diff
@@ -94,8 +96,6 @@ set(PROJECT_SOURCES
     textdocument_p.h
     texteditor.h
     texteditor.cpp
-    textlocation.h
-    textlocation.cpp
```
LeonMatthesKDAB, line 98: Why were they removed from CMake? The files still exist :thinking:

LeonMatthesKDAB (changes requested):
Please use references instead of pointers where appropriate. I also don't understand why this still builds without textlocation.cpp in CMake, but otherwise looks good.

src/core/CMakeLists.txt
narnaud: Oups, I was in the middle of something else...

src/core/lsp_utils.h
narnaud: In fact, this method can stay in CodeDocument, it doesn't add any LSP dependencies, and it makes sense to keep it there.

LeonMatthesKDAB: Ah, okay, the same question about pointer vs. reference also applies to lspFromPos, lspToPos and lspToRange.
Sorry, I forgot to mention that in the initial review.
These should all take TextDocument references I think instead of pointers.

narnaud: Done

LeonMatthesKDAB (approved):

PR #85, opened 2024-07-02
fix(build): Fix debug build on Windows

narnaud:
The -O2 compiler option is only available in release.


3rdparty/CMakeLists.txt
```diff
@@ -57,7 +57,7 @@ add_subdirectory(ksyntaxhighlighting)
 # * Enable optimizations even in debug mode (in a portable way)
 function(enable_optimizations target)
   if(MSVC)
-    target_compile_options(${target} PRIVATE /O2)
+    target_compile_options(${target} PRIVATE "$<$<CONFIG:Release>:/O2>")
```
dfaure-kdab, line 60: But then the whole line is useless. In release mode we're getting /O2 already, aren't we?

Sounds like

1. MSVC is stupid ;-)
2. this function should just do if (NOT MSVC) ...(-O3) endif() and leave MSVC untouched.

narnaud: You are correct, I missed the comment below...
You can't mix debug/release lib in Windows, it's jsut like that. With dlls, we see that at runtime, but here we are using static libs.

dfaure-kdab: I know you can't mix debug/release libs on Windows, but I thought this was about them linking to different CRT DLLs. I am VERY surprised that merely enabling optimizations in a debug build of a library is also not possible. Very strange. It must trigger more magic than just enabling optimizations then.

dfaure-kdab:
Change approved but I have no button to actually approve officially.

Issue #56, opened 2024-06-24
QML: Rename `Scripts` module


LeonMatthesKDAB:
The `Scripts` name is somewhat nondescript.

We should rename this to `Knut`, or `com.kdab.knut` before releasing v1.0.
The exact name still needs to be chosen.

narnaud:
I would prefer `import Knut`

Also take this opportunity to remove all version numbers in QML/JS
That's the new way to do things in Qt, so just `import Knut`, not `import Knut 1.0`

PR #94, opened 2024-07-09
refactor!: Rename Script QML module to Knut

LeonMatthesKDAB:
"import Script" was rather generic, so before release 1.0, we will want
to rename it to "import Knut".


narnaud (dismissed):
Just add `Fix #56` in the commit + update the submodules, and good to go.

LeonMatthesKDAB:
@narnaud should be good to go now, can you reapprove?

narnaud (approved):

PR #96, opened 2024-07-09
refactor!: Remove Knut.Test

### LINKED ISSUE #49: Remove Knut.Test all together
We don't need that anymore, there's a TestCase directly in Qt using QTest:
https://doc.qt.io/qt-6/qml-qttest-testcase.html

- [ ] refactor the test-data/tst_*.qml to use QTest
- [ ] move the static methods from core/TestUtil to test_utils.h (they are not used anymore in the QML files
- [ ] remove TestUtil and TestCase.qml from core

LeonMatthesKDAB:
We can replace this with QtTest in Qt6.

Fix #49


narnaud (commented):
Looks good.

Missing running cpp2doc (to remove the documentation for Knut.Test)... I'm starting to think we should get it back :)

I'll give it a shot on Windows to see why it's failing.

LeonMatthesKDAB:
That would be great! No clue why it's not working right now :thinking:

LeonMatthesKDAB:
After some discussions with @narnaud , we've decided to drop this PR.

While QtTest is very similar on the surface, it actually doesn't have an easy way to make sure all tests have run and to check which or even how many tests failed.
This also leads to some race conditions as the tests in QtTest are run by a Timer, which our ScriptRunner can't easily check for completion.

So let's abandon this and #49 for now.

PR #99, opened 2024-07-09
fix: add stdset="0" we don't want to generate property

Montel:


src/core/qtuidocument.cpp
```diff
@@ -263,15 +263,16 @@ QVariant QtUiWidget::getProperty(const QString &name) const
...
  *
  * Attributes is a has<string, string> object, where the key is the attribute name and the value is the attribute value.
  * For example:
  *
  * ```
  * widget.setProperty("text", "My text", { "comment": "some comment for translation" });
  * ```
+ * Set userProperty to true if you doesn't want to generate property.
  */
 void QtUiWidget::addProperty(const QString &name, const QVariant &value, const QHash<QString, QString> &attributes)
```
narnaud, line 277: Missing the new parameter here (and in the header) ;)

narnaud (changes requested):

src/core/qtuidocument.cpp
```diff
@@ -263,15 +263,16 @@ QVariant QtUiWidget::getProperty(const QString &name) const
 }
 
 /*!
- * \qmlmethod QtUiWidget::addProperty(string name, var value, object attributes = {})
- * Adds a new property with the given `name`, `value` and `attributes`.
+ * \qmlmethod QtUiWidget::addProperty(string name, var value, object attributes = {}, bool userProperty = false)
+ * Adds a new property with the given `name`, `value`, `attributes` and `useProperty`.
```
LeonMatthesKDAB, line 267: Is it `useProperty` or `userProperty`?

Montel: userProperty

narnaud (approved):

PR #100, opened 2024-07-09
chore(cmake): Re-enable docs target by default

### LINKED ISSUE #59: Refactor doc generation
Refactor the way we are generating the documentation. Right now, cpp2doc is called at every compilation, which is a waste of time, specially if there are no changes.

What we would like is:
- [x] remove cpp2doc from the ALL target (need explicit call)
- [ ] only call cpp2doc during a release, to ensure everything is up to date (doesn't prevent from updating the doc manually)

I wonder what would be the best way to handle that, but surely running cpp2doc every time is a waste of resources.

LeonMatthesKDAB:
Add a KNUT_UPDATE_DOCS option in CMake to optionally disable the
automatic documentation generation.
This is necessary as Qt Creator apparently fails running cpp2doc.

Fix #59

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,11 @@ option(USE_ASAN "use Address Sanitizer" OFF)
 # default.
 option(KNUT_ERROR_ON_WARN
        "Issue a compiler error if the compiler encounters a warning" OFF)
+option(
+  KNUT_UPDATE_DOCS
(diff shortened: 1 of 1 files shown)


LeonMatthesKDAB:
Oh, this is already #100 
These repos grow up so quickly :sweat_smile:

narnaud (approved):

PR #102, opened 2024-07-10
refactor!: Change the test API for scripts

narnaud:
This change generalize the `test` function for script, allowing to have multiple test functions.
Also remove the Knut.Test module, not needed anymore.

This makes the whole testing more streamline, and easier to work with.

- remove the `Knut.Test` module and `TestCase` item
- add `compare` and `verify` to `Script` and `ScriptDialog`
- run all methods starting with `test_` in test mode
- don't run `test` anymore (breaking change here), as it does not make sense to have a special case for this.

Related to #49


narnaud (commented):
The tests for mfc-utils won't pass, I'll fix it once we agree on the general direction here.

src/core/scriptrunner.cpp
```diff
@@ -205,6 +200,62 @@ bool ScriptRunner::isProperty(const QString &apiCall)
     return m_properties.contains(apiCall);
 }
 
+void ScriptRunner::compare(QObject *object, const QJSValue &actual, const QJSValue &expected, QString message)
```
LeonMatthesKDAB, line 203: Why do these functions live in the ScriptRunner and not just in the ScriptDialogItem.cpp?
Then they could also more easily set the failed member.

LeonMatthesKDAB (commented):
This actually looks really good :+1: 
It moves logic into C++, and get's rid of 180 lines of code. What's not to like?

I'm just wondering whether some of the methods should live in the ScriptDialogItem directly, but that's minor nitpicks at this point.

src/core/scriptrunner.cpp
LeonMatthesKDAB: Ah, okay, that makes sense :+1:

LeonMatthesKDAB (approved):

PR #104, opened 2024-07-10
fix: allow to replace by empty string (=> remove text)

Montel:


narnaud (approved):
Nice catch, didn't think of it.

PR #105, opened 2024-07-10
feat: Select Smaller/Larger/Next/Previous Syntax Nodes

### LINKED ISSUE #64: Add "select larger syntax node" and "select smaller syntax node"
In the C++ menu, we have something call Select Block Up... It would be nice to move that to `CodeDocument`, and implement it with TreeSitter with "select larger syntax node" and "select smaller syntax node".

That would make it way more useful.

Ideas coming from this blog: https://zed.dev/blog/syntax-aware-editing

LeonMatthesKDAB:
This allows navigating the code semantically.

Fix #64

### DIFF
--- a/src/core/codedocument.cpp
+++ b/src/core/codedocument.cpp
@@ -405,6 +405,74 @@ void CodeDocument::selectSymbol(const QString &name, int options)
 }
 
 /*!
+ * \qmlmethod int CodeDocument::selectLargerSyntaxNode(int count = 1)
+ *
+ * Selects the text of the next larger syntax node that the selection is in.
+ *
+ * It does so `count` times and returns the resulting cursor position.
+ */
+int CodeDocument::selectLargerSyntaxNode(int count /* = 1*/)
+{
+    LOG("CodeDocument::selectLargerSyntaxNode", LOG_ARG("count", count));
+
+    auto currentNode = m_treeSitterHelper->nodeCoveringRange(selectionStart(), selectionEnd());
+
+    auto matchesCurrentSelection = static_cast<int>(currentNode.startPosition()) == selectionStart()
+        && static_cast<int>(currentNode.endPosition()) == selectionEnd();
+    if (!matchesCurrentSelection) {
+        // finding the node that covers the current selection already produced a larger syntax node.
+        // So we're already up one level.
+        --count;
+    }
+
+    for (/*count already initialized*/; count > 0 && !currentNode.parent().isNull(); --count) {
+        auto largerNode = currentNode.parent();
+        if (largerNode.startPosition() == currentNode.startPosition()
+            && largerNode.endPosition() == currentNode.endPosition()) {
+            // If the parent node matches the current selection exactly, that's not a real change, so search one
+            // additional time.
+            ++count;
+        }
+        currentNode = largerNode;
+    }
+
+    selectRegion(currentNode.startPosition(), currentNode.endPosition());
+
+    return position();
+}
+
+int CodeDocument::selectSmallerSyntaxNode(int count /* = 1*/)
+{
+    LOG("CodeDocument::selectSmallerSyntaxNode", LOG_ARG("count", count));
+
+    auto smallerNodes =
+        kdalgorithms::filtered(m_treeSitterHelper->nodesInRange(createRangeMark()), [](const auto &node) {
+            return node.isNamed();
+        });
+
+    std::optional<treesitter::Node> node;
+    for (/*count already initialized*/; count > 0; --count) {
+        if (smallerNodes.isEmpty()) {
+            break;
+        }
+        node = smallerNodes.first();
+        auto matchesCurrentSelection = static_cast<int>(node->startPosition()) == selectionStart()
+            && static_cast<int>(node->endPosition()) == selectionEnd();
+        // If the first found node matches the current selection exactly, we need to search one additional time.
+        if (matchesCurrentSelection) {
+            ++count;
+        }
+        smallerNodes = node->namedChildren();
+    }
+
+    if (node.has_value()) {
+        selectRegion(node->startPosition(), node->endPosition());
+    }
+
+    return position();
+}
+
+/*!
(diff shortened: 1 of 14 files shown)


src/core/codedocument.cpp
```diff
@@ -404,6 +404,160 @@ void CodeDocument::selectSymbol(const QString &name, int options)
...
+            ++count;
+        }
+        smallerNodes = node->namedChildren();
+    }
+
+    if (node.has_value()) {
+        selectRegion(node->startPosition(), node->endPosition());
+    }
+
+    return position();
```
narnaud, line 481: Same here `LOG_RETURN("pos", position());`

src/core/codedocument.cpp
```diff
@@ -404,6 +404,160 @@ void CodeDocument::selectSymbol(const QString &name, int options)
...
+ * Selects the left-most next smaller syntax node within the current selection.
+ *
+ * It does so `count` times and returns the resulting cursor position.
+ *
+ * Note that this only selects "named" Tree-sitter nodes, so punctuation and other unnamed nodes are skipped.
+ */
+int CodeDocument::selectSmallerSyntaxNode(int count /* = 1*/)
+{
+    LOG_AND_MERGE("CodeDocument::selectSmallerSyntaxNode", LOG_ARG("count", count));
+
```
narnaud, line 456: We should probably show a info and stop here if there a no selection.
When testing, I was puzzled by the warning I got:
```
[2024-07-11 20:05:01.730] [warning] TextDocument::createRangeMark: Creating a range mark with an empty range.
```

src/core/codedocument.cpp
```diff
@@ -404,6 +404,160 @@ void CodeDocument::selectSymbol(const QString &name, int options)
...
+    auto node = m_treeSitterHelper->nodeCoveringRange(selectionStart(), selectionEnd());
+
+    auto target = findSibling(node, &treesitter::Node::nextNamedSibling, count);
+
+    if (target.has_value()) {
+        selectRegion(target->startPosition(), target->endPosition());
+    }
+
+    return position();
```
narnaud, line 531: Same here `LOG_RETURN("pos", position());`

src/gui/mainwindow.h
```diff
@@ -67,19 +67,24 @@ class MainWindow : public QMainWindow
...
     // C++
-    void followSymbol();
-    void switchDeclarationDefinition();
-    void switchHeaderSource();
+    void openHeaderSource();
     void gotoBlockStart();
     void gotoBlockEnd();
     void selectBlockStart();
     void selectBlockEnd();
-    void selectBlockUp();
```
narnaud, line 77: Please keep it, even if it's doable with selectLargerNode, it's slightly different enough to make sense to keep it around.

Shortcut is Ctrl+Shift+U, should be ok

src/gui/mainwindow.cpp
```diff
@@ -632,9 +586,15 @@ void MainWindow::updateActions()
     ui->actionGotoBlockStart->setEnabled(cppEnabled);
     ui->actionSelectToBlockEnd->setEnabled(cppEnabled);
     ui->actionSelectToBlockStart->setEnabled(cppEnabled);
-    ui->actionSelectBlockUp->setEnabled(cppEnabled);
     ui->actionDeleteMethod->setEnabled(cppEnabled);
 
+    const bool isCodeDocument = codeDocument != nullptr;
+    ui->actionSelectLargerSyntaxNode->setEnabled(isCodeDocument);
+    ui->actionSelectSmallerSyntaxNode->setEnabled(isCodeDocument);
```
narnaud, line 593: Would be nice to disable if there are no selections, as it does nothing.

src/gui/mainwindow.cpp
```diff
@@ -275,53 +278,31 @@ bool MainWindow::eventFilter(QObject *watched, QEvent *event)
...
-    if (auto cppDocument = qobject_cast<Core::CppDocument *>(Core::Project::instance()->currentDocument()))
-        cppDocument->gotoBlockStart();
-}
-
-void MainWindow::gotoBlockEnd()
-{
-    if (auto cppDocument = qobject_cast<Core::CppDocument *>(Core::Project::instance()->currentDocument()))
-        cppDocument->gotoBlockEnd();
-}
+#define FORWARD_TO_DOCUMENT(TYPE, FUNCTION)                                                                            \
```
narnaud, line 281: Nice, can you move before the constructor, and add a bit of documentation on the macros?

The reason is in case someone add a method before in the header file, and create the definition from the IDE, it will most likely be created before the macro.

I think you can also use it for selectAll below.

narnaud (changes requested):
Very nice, it's actually working nicely.
The previous/next sibling can be really useful for fast scripting.

Small comments, nothing serious.

src/gui/mainwindow.cpp
LeonMatthesKDAB: Done

src/gui/mainwindow.h
LeonMatthesKDAB: Re-added it

src/core/codedocument.cpp
LeonMatthesKDAB: Done

src/core/codedocument.cpp
```diff
@@ -404,6 +404,160 @@ void CodeDocument::selectSymbol(const QString &name, int options)
...
+            // If the parent node matches the current selection exactly, that's not a real change, so search one
+            // additional time.
+            ++count;
+        }
+        currentNode = largerNode;
+    }
+
+    selectRegion(currentNode.startPosition(), currentNode.endPosition());
+
+    return position();
```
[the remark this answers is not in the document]
LeonMatthesKDAB: Done for all 4 methods

narnaud (approved):

src/gui/mainwindow.cpp
narnaud: We have the warning, probably enough.
We can add a task for later, but it's probably good enough.

PR #106, opened 2024-07-12
fix: save settings file in different directory in testing mode

smnppKDAB:
before this commit, we didn't save json files in test mode, so as not to overwrite the json file already present in the user's home directory. now we save this json file in the temp directory.

We have also improved tst_settings.cpp, which now saves the json file in the temp directory.

### DIFF
--- a/src/core/settings.cpp
+++ b/src/core/settings.cpp
@@ -231,6 +231,9 @@ bool Settings::setValue(QString path, const QVariant &value)
 
 QString Settings::userFilePath() const
 {
+    if (isTesting()) {
+        return QDir::tempPath() + '/' + SettingsName;
+    }
     return QDir::homePath() + '/' + SettingsName;
 }
 
@@ -268,10 +271,6 @@ void Settings::loadKnutSettings()
 
 void Settings::saveSettings()
 {
-    // Don't save settings if testing
-    if (isTesting())
-        return;
-
     const auto &settings = isUser() ? m_userSettings : m_projectSettings;
     const auto &filePath = isUser() ? userFilePath() : projectFilePath();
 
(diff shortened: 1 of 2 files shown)


src/core/settings.cpp
```diff
@@ -231,6 +231,9 @@ bool Settings::setValue(QString path, const QVariant &value)
 
 QString Settings::userFilePath() const
 {
+    if (isTesting()) {
+        return QDir::tempPath() + '/' + SettingsName;
+    }
```
narnaud, line 236: This won't help. The idea is to ditch all user settings in test mode.
Imagine that you change a setting in a test, and later on another test is using the same settings: you create a dependency between the 2 tests.

src/core/settings.cpp
```diff
@@ -268,10 +271,6 @@ void Settings::loadKnutSettings()
 
 void Settings::saveSettings()
 {
-    // Don't save settings if testing
-    if (isTesting())
```
narnaud, line 272: In fact, I think the right fix is just to allow saving project settings, and not user settings.
So `if (isTesting() && isUser()) return;`

narnaud (changes requested):

src/core/settings.cpp
dfaure-kdab: That's exactly what happens, because reading the settings doesn't happen in Test mode, see the constructor of Settings.

narnaud: Oh! forgot about that... you're right.

src/core/settings.cpp
narnaud: Nevermind, as you can only write the user settings in test mode.

narnaud (approved):

PR #108, opened 2024-07-15
Remove TextLocation and TextRange class

### LINKED ISSUE #67: Refactor the text range/mark API

narnaud:
Replaced with the existing MarkRange.

The rename, if any, is left out and will be part of #54.

Fix #67


knut.code-workspace
```diff
@@ -17,7 +17,91 @@
...
+            "xlocale": "cpp",
+            "xlocbuf": "cpp",
+            "xlocinfo": "cpp",
+            "xlocmes": "cpp",
+            "xlocmon": "cpp",
+            "xlocnum": "cpp",
+            "xloctime": "cpp",
+            "xmemory": "cpp",
+            "xtr1common": "cpp",
+            "xtree": "cpp"
```
LeonMatthesKDAB, line 104: This seems to be half of STL? Is that on purpose?

LeonMatthesKDAB (commented):
Apart from the change to the VS Code Workspace file this looks good :+1:

knut.code-workspace
narnaud: No, my fault, I missed it. I need to fix that in a different commit... That's VSCode C++ extension adding this.

LeonMatthesKDAB (dismissed):

PR #114, opened 2024-07-16
Add a split view for code document with TreeSitter

narnaud:
The right view shows the TreeSitter tree, like in the inspector view.


src/gui/gui/file-tree.png
LeonMatthesKDAB, line 1: What's the origin of this image and what's the license?

src/gui/mainwindow.cpp
```diff
@@ -717,13 +718,19 @@ static QWidget *widgetForDocument(Core::Document *document)
         tsView->setTsDocument(qobject_cast<Core::QtTsDocument *>(document));
         return tsView;
     }
-    case Core::Document::Type::Json:
     case Core::Document::Type::Cpp:
+    case Core::Document::Type::Json:
     case Core::Document::Type::Text:
     default: {
```
LeonMatthesKDAB, line 724: Why is CodeView and TextView put into the same `case` here, but all others are separated by case?

Why is it not:
```cpp
case Document::Type::Cpp:
     auto codeView = new CodeView();
    codeView->setDocument(qobject_cast<Core::CodeDocument*>(codeDocument));
    return codeView;
```
And keep the default as-is?

src/gui/codeview.h
```diff
@@ -0,0 +1,52 @@
...
+    void initializeCodeModel();
+    void changeText();
+    void changeCursor();
+    void changeTreeSelection(const QModelIndex &current);
+    std::unique_ptr<treesitter::Predicates> makePredicates();
+
+    QTreeView *m_treeView = nullptr;
+    QCheckBox *m_showUnnamed = nullptr;
+    treesitter::Parser m_parser;
+    TreeSitterTreeModel *m_treemodel = nullptr;
```
LeonMatthesKDAB, line 49: Can we share this model with the TreeSitterInspector?
I'd like to keep the inspector, as I actually quite like having it in a separate window, and it still provides the queries, which this doesn't (yet?) have.
But then it would make sense to only have a single model that's shared in order to save on resources. Updating these models isn't quite trivial.

LeonMatthesKDAB (changes requested):
The resulting UI looks good.

However, we should consider the performance impact of this change.
The TreeSitterTreeModel is now used twice and the one on the code model will always be kept up to date, even if the view isn't visible.
Especially with large documents, this could become slow.

Also, there's a lot of code duplication with the TreeSitterInspector. I'm in favor of keeping it as a separate window as well, but can we at least share the models?

src/gui/mainwindow.cpp
narnaud: It was mostly to handle the default case, as we want a CodeView if the class inherit from CodeDocument.
But I suppose it doesn't make much sense, we should just have default be TextView and handle the rest independently.
It's not like we are going to add new views everyday.

src/gui/codeview.h
narnaud: Ok, my idea was to remove the TreeSitterInspector completely. But if we keep it, this doesn't make much sense.
So instead of integrating the treeview here, maybe we can keep the button (for discoverability) and open the inspector (or raise it if it's already open).
I don't see the point in having the tree in 2 different places.

So let's do that rather than my patch.

src/gui/gui/file-tree.png
narnaud: All icons are coming from here:
https://pictogrammers.com/docs/general/license/

Guess I need to document that somewhere...

narnaud:
Let's close it... doesn't make much sense if we keep the inspector as a second window.

PR #120, opened 2024-07-19
build: Fix include issue with TreeSitter and unicode

mmennuti-kdab:


narnaud (changes requested):
Please make sure to install pre-commit, see here:
https://kdab.github.io/knut/contributing/getting-involved.html#pre-commit

Particularly, we are using https://www.conventionalcommits.org/en/v1.0.0/, so you're commit message should follow the convention (strange that it's not found during the pre-commit check).

Also, a commit message should focus on the why rather than the what.

Something like:
```
build: Fix include issue with TreeSitter and unicode
```

LeonMatthesKDAB (approved):

LeonMatthesKDAB (approved):

PR #121, opened 2024-07-22
feat: add a method to change the context of a message

smnppKDAB:


src/core/qttsdocument.h
```diff
@@ -89,6 +89,9 @@ class QtTsDocument : public Document
     void initializeXml();
     pugi::xml_document m_document;
 
+    pugi::xml_node findContext(const QString &context);
```
dfaure-kdab, line 92: This method should be `const` (but only this one, not `findOrCreateContext`, obviously).

dfaure-kdab:
Using the IDS_FOO which is stored in the "comment" to look up a specific message was my idea, but looking back at the merge request now I have to wonder if that's the right way to do this. When knut is used in another context than a MFC migration, the comment might not be a unique identifier for the message.... I think we need to pass context+comment+source (the message to be translated), to better identify the message that we want to move (same "key" as what tr() uses, AFAIK).

docs/API/knut/qttsdocument.md
```diff
@@ -20,6 +20,7 @@ import Knut
 |-|-|
 ||**[addMessage](#addMessage)**(string context, string location, string source, string translation)|
 ||**[setLanguage](#setLanguage)**(string lang)|
+||**[setMessageContext](#setMessageContext)**(string comment, string context)|
```
dfaure-kdab, line 23: missing `source` (but see below, it will also be missing `newContext`)

src/core/qttsdocument.cpp
```diff
@@ -160,6 +160,70 @@ void QtTsDocument::addMessage(const QString &context, const QString &fileName, c
     Q_EMIT fileUpdated();
 }
 
+/*!
+ * \qmlmethod QtTsDocument::setMessageContext(string comment, string context)
+ * Set a new `context`for the message given with the `comment` as id
```
dfaure-kdab, line 165: Needs an update

src/core/qttsdocument.cpp
```diff
@@ -160,6 +160,70 @@ void QtTsDocument::addMessage(const QString &context, const QString &fileName, c
...
+ */
+
+void QtTsDocument::setMessageContext(const QString &comment, const QString &source, const QString &context)
```
dfaure-kdab, line 168: The key to identify a message is three things: context, comment, source.

So this method needs to take 4 parameters:

context, comment, source, newContext

The first three to identify the message to be changed, and the last one to define the new context to move it to.

dfaure-kdab (changes requested):

docs/API/knut/qttsdocument.md
```diff
@@ -46,6 +47,10 @@ Adds a new `source` text, its `translation` located in `location` within the giv
 
 Changes the language.
 
+#### <a name="setMessageContext"></a>**setMessageContext**(string context, string comment, string source, string newContext)
+
+Set a `newContext` for the message given with the `context`, `comment` and `source` strings
```
dfaure-kdab, line 52: given with -> identified by

dfaure-kdab (approved):

PR #122, opened 2024-07-22
feat: added line property for a message node

smnppKDAB:


src/core/qttsdocument.cpp
```diff
@@ -266,6 +266,19 @@ void QtTsMessage::setFileName(const QString &file)
...
 }
 
+QString QtTsMessage::line() const
+{
+    return QString::fromLatin1(m_message.child("location").attribute("line").value());
+}
+
+void QtTsMessage::setLine(const QString &line)
+{
+    LOG("QtTsMessage::setFileName", line);
```
dfaure-kdab, line 276: This isn't setFileName (copy/paste error).

src/core/qttsdocument.h
```diff
@@ -22,6 +22,7 @@ class QtTsMessage : public QObject
     Q_OBJECT
     Q_PROPERTY(QString fileName READ fileName WRITE setFileName NOTIFY fileNameChanged)
     Q_PROPERTY(QString source READ source WRITE setSource NOTIFY sourceChanged)
+    Q_PROPERTY(QString line READ line WRITE setLine NOTIFY lineChanged)
```
dfaure-kdab, line 25: Why not make it an `int` instead? At least C++ wise where types matter, an int seems more logical for a line number.

dfaure-kdab (approved):

PR #125, opened 2024-07-24
fix: save settings on exit

smnppKDAB:
We add a way to save settings before exiting if there are changes. This make the tests to pass with the corrected knut.json

### DIFF
--- a/src/core/settings.h
+++ b/src/core/settings.h
@@ -136,6 +136,7 @@ private:
     void updatePaths(const QString &path, const std::string &json_path, bool add);
     void saveSettings();
     bool isUser() const;
+    void saveOnExit();
 
     inline static Settings *m_instance = nullptr;
 
--- a/src/core/knutcore.cpp
+++ b/src/core/knutcore.cpp
@@ -86,6 +86,10 @@ void KnutCore::process(const QStringList &arguments)
     else
         mode = Settings::Mode::Gui;
 
+    // Finish scripts using scriptFinished(), not just when the last window is closed
+    if (mode == Settings::Mode::Test)
+        QApplication::setQuitOnLastWindowClosed(false);
+
     initialize(mode);
 
     const QStringList positionalArguments = parser.positionalArguments();
--- a/src/core/settings.cpp
+++ b/src/core/settings.cpp
@@ -85,6 +85,7 @@ Settings::Settings(Mode mode, QObject *parent)
 
 Settings::~Settings()
 {
+    saveOnExit();
     m_instance = nullptr;
 }
 
@@ -296,6 +297,13 @@ void Settings::saveSettings()
     QTextStream stream(&file);
     stream << QString::fromStdString(settings.dump(4, ' ', false));
     emit settingsSaved();
+}
+
+void Settings::saveOnExit()
+{
+    if (m_saveTimer->isActive()) {
+        saveSettings();
+    }
 }
 
 bool Settings::isUser() const


dfaure-kdab:
Looks good to me, but this is my idea in the first place, so I'll let @LeonMatthesKDAB review ;-)

Debugging indicated that with offscreen QPA the Settings timer had time to fire but without it, the app would exit due to lastWindowClosed(), without waiting for scriptFinished and without saving settings.

LeonMatthesKDAB (changes requested):
Overall making sure we save the settings when the timer is still active definitely makes sense.
However, can you try calling saveOnExit() in the Settings destructor, instead of connecting it to scriptFinished?

It just seems like that would be the more fool-proof option and should cover more cases.

LeonMatthesKDAB:
Hm, I just tried testing knut locally and this PR actually doesn't seem to fix the issue :thinking: 
Strangely 2/3 tests fail without actually telling me which file is different.
And the other one tells me that knut.json is missing in the `expected` folder.
Which to me almost sounds like it's now saving settings that it didn't used to before.

LeonMatthesKDAB:
So for the mfc_convert_ribbon test I get the following files:

expected
```
{
    "qtui": {
        "custom_widgets": [
            "MySliderCtrl|\"mysliderctrl.h\"|QSlider"
        ]
    }
}
```

actual
```
{
    "qtui": {
        "custom_widgets": [
            "MySliderCtrl|\"mysliderctrl.h\"|QSlider"
        ]
    },
    "rc": {
        "dialog_flags": [
            "UpdateGeometry",
            "UseIdForPixmap"
        ],
        "dialog_scalex": 1.5,
        "dialog_scaley": 1.65
    }
}
```

So is this change correct?

dfaure-kdab:
Don't forget the additional variable of running tests via ctest/CI versus running tests directly (QtCreator or command-line).
This patch here is about ensuring that both leads to the same result. Previously, ctest/CI would save on exit while QtCreator/cmdline would not (because of lastWindowClosed).

So yes, if you stay within the ctest/CI use case, this patch isn't needed. But better save in all cases to minimize surprises.

Calling saveOnExit() in the Settings destructor is an interesting idea, I'll let @smnppKDAB test it.

smnppKDAB:
> Hm, so for me the tests only pass with these patches: https://codereview.kdab.com/c/kdab/mfc-utils/+/144535
> 
> Are these updates expected? I'm not too familiar with mfc-utils, so can't tell if they are sensible.

Here is the corresponding MR : https://codereview.kdab.com/c/kdab/mfc-utils/+/144526

src/core/settings.h
```diff
@@ -104,6 +104,8 @@ class Settings : public QObject
     Q_INVOKABLE bool hasValue(QString path) const;
     Q_INVOKABLE QVariant value(QString path, const QVariant &defaultValue = {}) const;
 
+    void saveOnExit();
```
dfaure-kdab, line 107: Should be private now that it's only called internally.

LeonMatthesKDAB:
> > Hm, so for me the tests only pass with these patches: https://codereview.kdab.com/c/kdab/mfc-utils/+/144535
> > Are these updates expected? I'm not too familiar with mfc-utils, so can't tell if they are sensible.
> 
> Here is the corresponding MR : https://codereview.kdab.com/c/kdab/mfc-utils/+/144526

Oh, right, I didn't know about that MR. Please link any MRs that a PR is related to next time :)

src/core/settings.h
LeonMatthesKDAB: Why has this been resolved? It is still public.

LeonMatthesKDAB (approved):

dfaure-kdab (approved):

Issue #72, opened 2024-06-28
Improve nextStep API in ScriptDialog


narnaud:
The `nextStep` API is good for big independent steps, but bad showing progress of one action (for example, the same action being done on 1000 items).

It would be good to improve the API:
- if the string is empty (ie calling `nextStep()`), then repeat the previous string
- don't update more than 5x times per seconds (meaning you need a timer to do the update).

Would be good to have some numbers to see how it impacts script with a before/after.

smnppKDAB:
Fix in the PR#129

PR #129, opened 2024-07-25
feat: improve nextStep API in ScriptDialog

smnppKDAB:
- if the string is empty (ie calling nextStep()), then repeat the previous string
 - Depends on : https://codereview.kdab.com/c/kdab/mfc-utils/+/144576


dfaure-kdab (approved):

smnppKDAB:
this fixes the issue #72

PR #130, opened 2024-07-26
fix: improve script performance by fixing size of row in history panel

smnppKDAB:
this patch reduces the running time of scripts like mfc_update_translation.qml from 80000ms to about 500ms


dfaure-kdab:
+1 from me (those were my suggestions after watching the profiler output on Simon's Mac)

The x160 speedup is pretty impressive.

PR #131, opened 2024-07-29
fix: Make Symbols & query... calls consistent

### LINKED ISSUE #118: Check consistency between `CodeDocument::symbols` and `CppDocument::*Query`
Make sure that we are consistent in the data we are returning between the symbols and the *Query methods from CppDocument.

LeonMatthesKDAB:
We have a lot of symbols that have corresponding functions to query them directly.
Before this commit they actually used completely different queries, which could result in inconsistency between the data returned from a symbol and from the query that should query the same thing.

This also fixes many pointer captures that weren't correct in many instances.

Fix #118


narnaud (approved):
Very nice!

PR #132, opened 2024-07-29
fix: simplify the API documentation

smnppKDAB:
Fixes KDAB#50

This is a follow up of #49.
- remove all "\inqmlmodule in the code documentation
- add an empty string into Script module to be able to use it, just in case for later

### DIFF
--- a/tools/cpp2doc/sourceparser.cpp
+++ b/tools/cpp2doc/sourceparser.cpp
@@ -103,6 +103,10 @@ Data::TypeBlock SourceParser::parseType(QTextStream &stream, QString line)
             parseBlock(line, currentType);
     }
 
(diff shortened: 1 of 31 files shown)


tools/cpp2doc/sourceparser.cpp
```diff
@@ -103,6 +103,10 @@ Data::TypeBlock SourceParser::parseType(QTextStream &stream, QString line)
             parseBlock(line, currentType);
     }
 
+    if (currentType.qmlModule.isEmpty()) {
+        currentType.qmlModule = "Script";
```
narnaud, line 107: The module name has changed to Knut in-between (ie after I created the issue).
So change `"Script"` for `"Knut"` here.

dfaure-kdab (approved):

Issue #62, opened 2024-06-24
Improve the Tree-Sitter inspector


narnaud:
More discussion, quite some changes as we did a 180° turn.

- we keep the inspector as a second window
- we add a button for the code views to open the inspector (for better discoverability)
- improvement on the query editor
    - better syntax highlighting (use scm for that)
    - load/save queries (using scm extension for that)
    - wiggle on error line

narnaud:
Here is a proposal:
- Create a CodeView that could be split, with the code on the left and the TreeSitter tree on the right (one button click will show/hide it), something like [Neovim TreeSitter](https://user-images.githubusercontent.com/2361214/202389106-244ac890-9442-4759-9b2c-4fe3c247dfbc.gif)
- Add a dock for writing TreeSitter query, that you could save and load (apparently the extension is .scm)

This is completely removing the current dialog, but better integrate TreeSitter into Knut.

LeonMatthesKDAB:
Sounds good overall. The inspector could definitely use some love. It's been working fine but it's UX isn't great right now.

Side Note: .scm stands for scheme. Tree-sitter doesn't actually use scheme syntax, but it's close enough that highlighting, etc. should work fine.
We should check whether KSyntaxHighlighter has a lisp/scheme highlighter, which could be good enough.

PR #133, opened 2024-07-31
feat: Scheme syntax highlighting for TreeSitter

LeonMatthesKDAB:
This uses KSyntaxHighlighting::SyntaxHighlighter to enable a
scheme-based syntax highlighting on the TreeSitterInspector.

This still has a few issues, as e.g. #eq? produces a strange
highlighting result and certain keywords are marked up without reason.
But it's still a good improvement from no syntax highlighting.

Part of #62.


itzurabhi (approved):
LGTM

PR #134, opened 2024-07-31
fix!: Remove TreeSitter transformations

LeonMatthesKDAB:
TreeSitter Transformations were unused but remained in the codebase, even though JavaScript string interpolation
is the better way to go anyway.

So remove this feature entirely before 1.0 so we don't have to support it.

Based on #133
Related to #62


itzurabhi (approved):
LGTM

PR #136, opened 2024-07-31
feat: add a way to initialize script dialog from command line

smnppKDAB:
Fixes KDAB#32
Tested in de MR : https://codereview.kdab.com/c/kdab/mfc-utils/+/144788

### DIFF
--- a/src/core/knutcore.cpp
+++ b/src/core/knutcore.cpp
@@ -124,6 +124,18 @@ bool KnutCore::process(const QStringList &arguments)
         }
     }
 
+    // Get json data if provided
+    const QString jsonDataStr = parser.value("data");
+    json jsonData;
+    if (!jsonDataStr.isEmpty()) {
+        try {
+            jsonData = json::parse(jsonDataStr.toStdString());
+        } catch (const json::parse_error &ex) {
+            spdlog::error("JSON parsing error at byte {}: {}", ex.byte, ex.what());
+            return false;
+        }
+    }
+
     // Run the script passed in parameter, if any
     // Exit Knut if there are no windows opened
     auto scriptName = parser.value("run");
@@ -132,8 +144,8 @@ bool KnutCore::process(const QStringList &arguments)
     }
 
     if (!scriptName.isEmpty()) {
-        QTimer::singleShot(0, this, [scriptName]() {
-            ScriptManager::instance()->runScript(scriptName);
+        QTimer::singleShot(0, this, [scriptName, jsonData = std::move(jsonData)]() mutable {
+            ScriptManager::instance()->runScript(scriptName, std::move(jsonData));
         });
         connect(
             ScriptManager::instance(), &ScriptManager::scriptFinished, qApp,
@@ -160,6 +172,7 @@ void KnutCore::initParser(QCommandLineParser &parser) const
                        {{"i", "input"}, "Opens document <file> on startup.", "file"},
                        {{"l", "line"}, "Line in the current file, if any.", "line"},
                        {{"c", "column"}, "Column in the current file, if any.", "column"},
+                       {{"d", "data"}, "JSON data string for initializing the dialog.", "data"},
                        {"json-list", "Returns the list of all available scripts as a JSON file"},
                        {"json-settings", "Returns the settings as a JSON file"}});
 }
--- a/src/core/scriptdialogitem.cpp
+++ b/src/core/scriptdialogitem.cpp
@@ -24,6 +24,9 @@
 #include <QDoubleSpinBox>
 #include <QFile>
 #include <QFileInfo>
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <QJsonValue>
 #include <QLineEdit>
 #include <QMessageBox>
 #include <QPushButton>
@@ -143,6 +146,29 @@ ScriptDialogItem::ScriptDialogItem(QWidget *parent)
     m_data->registerDataChangedCallback([this](const QString &key, const QVariant &value) {
         changeValue(key, value);
     });
+}
+
+void ScriptDialogItem::initialize(nlohmann::json &&jsonData)
+{
+    for (auto it = jsonData.cbegin(); it != jsonData.cend(); ++it) {
+        const QString key = QString::fromStdString(it.key());
+
+        QVariant value;
+        if (it.value().is_string()) {
+            value = QString::fromStdString(it.value().get<std::string>());
+        } else if (it.value().is_boolean()) {
+            value = it.value().get<bool>();
+        } else if (it.value().is_number_integer()) {
+            value = it.value().get<int>();
+        } else if (it.value().is_number_float()) {
+            value = it.value().get<double>();
+        } else {
+            spdlog::error("Unsupported data type for key '{}'", it.key());
+            value = QString::fromStdString(it.value().dump());
+        }
+
+        changeValue(key, value);
+    }
 }
 
 void ScriptDialogItem::done(int code)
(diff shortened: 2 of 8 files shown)


src/core/knutcore.cpp
```diff
@@ -132,8 +139,8 @@ bool KnutCore::process(const QStringList &arguments)
     }
 
     if (!scriptName.isEmpty()) {
-        QTimer::singleShot(0, this, [scriptName]() {
-            ScriptManager::instance()->runScript(scriptName);
+        QTimer::singleShot(0, this, [scriptName, jsonData]() {
```
LeonMatthesKDAB, line 142:
```suggestion
        QTimer::singleShot(0, this, [scriptName, jsonData=move(jsonData)]() {
```

src/core/knutcore.cpp
```diff
@@ -160,6 +167,7 @@ void KnutCore::initParser(QCommandLineParser &parser) const
                        {{"i", "input"}, "Opens document <file> on startup.", "file"},
                        {{"l", "line"}, "Line in the current file, if any.", "line"},
                        {{"c", "column"}, "Column in the current file, if any.", "column"},
+                       {{"d", "data"}, "JSON data string or file for initializing the dialog.", "data"},
```
LeonMatthesKDAB, line 170: It says here that this parameter could point to a JSON file. This sounds very useful, but I think it's not actually implemented?

src/core/scriptdialogitem.cpp
```diff
@@ -145,6 +148,28 @@ ScriptDialogItem::ScriptDialogItem(QWidget *parent)
...
+            value = it.value().get<bool>();
+        } else if (it.value().is_number_integer()) {
+            value = it.value().get<int>();
+        } else if (it.value().is_number_float()) {
+            value = it.value().get<double>();
+        } else {
+            value = QString::fromStdString(it.value().dump());
+        }
+
+        changeValue(key, value);
```
LeonMatthesKDAB, line 169: As `changeValue` may now be called by user input, we will need proper error handling for it.
Please at least add a `spdlog::warning` calls to `changeValue` in case no widget could be found for the key.
Otherwise it could be quite frustrating to use if you don't even get notified about a typo in the key.

src/core/scriptdialogitem.h
```diff
@@ -14,8 +14,11 @@
 #include <QJSValue>
 #include <QQmlListProperty>
 #include <QQmlPropertyMap>
+#include <nlohmann/json.hpp>
 #include <vector>
 
+using json = nlohmann::json;
```
LeonMatthesKDAB, line 20: Don't add using directives in a header. All dependents will have the same `use` statement, which could lead to conflicts.
Prefer fully-qualifying in the header.

src/core/scriptmanager.cpp
```diff
@@ -77,7 +77,7 @@ QAbstractItemModel *ScriptManager::model()
     return model;
 }
 
-void ScriptManager::runScript(const QString &fileName, bool async, bool log)
+void ScriptManager::runScript(const QString &fileName, const json &data, bool async, bool log)
```
LeonMatthesKDAB, line 80: Ideally take the data by r-value reference so you can move it instead of copying.
It's user input that we don't know the size of, so copying may be expensive.

src/core/scriptrunner.cpp
```diff
@@ -289,6 +289,16 @@ QQmlEngine *ScriptRunner::getEngine(const QString &fileName)
...
+json ScriptRunner::jsonData()
+{
+    return m_jsonData;
+}
+
+void ScriptRunner::setJsonData(const json &data)
+{
+    m_jsonData = data;
+}
+
```
LeonMatthesKDAB, line 301: These functions are never used I think.
The member can probably go as well?

src/core/knutcore.cpp
```diff
@@ -124,6 +124,13 @@ bool KnutCore::process(const QStringList &arguments)
         }
     }
 
+    // Get json data if provided
+    const QString jsonDataStr = parser.value("data");
+    json jsonData;
+    if (!jsonDataStr.isEmpty()) {
+        jsonData = json::parse(jsonDataStr.toStdString());
```
LeonMatthesKDAB, line 131: This may throw an exception that we should handle gracefully.
See also: https://json.nlohmann.me/features/parsing/parse_exceptions/

LeonMatthesKDAB (dismissed):
Cool change overall, and an important feature :partying_face: 

There's still some loose ends to tidy up before we can merge this though.

LeonMatthesKDAB:
Oh, also side note: This should be `feat:` not `fix:`, it's adding a new feature, not fixing anything existing IMO :)

src/core/scriptdialogitem.cpp
```diff
@@ -535,6 +560,12 @@ void ScriptDialogItem::createProperties(QWidget *dialogWidget)
 void ScriptDialogItem::changeValue(const QString &key, const QVariant &value)
 {
     auto widget = findChild<QWidget *>(key);
+
+    if (!widget) {
```
LeonMatthesKDAB, line 564: With this change we won't ever reach the special case about Combobox Models.
It may be easier to just `return` whenever we found a value and warn at the end of the function instead?

As this warning also won't work in case that the widget exists but is of an unsupported type.

src/core/scriptmanager.cpp
LeonMatthesKDAB: This isn't using r-value references correctly.

1. const r-value references will force another copy (see: https://www.sandordargo.com/blog/2021/08/18/const-rvalue-references )
2. This function is declared in the header as taking an l-value reference
3. In order to avoid copies we would have to use r-value references all the way down from start to finish. There's still other functions that this is passed to/from that will force a copy of the data.

Maybe it's better if we stick with l-value references for now?

LeonMatthesKDAB (commented):
Still a few issues left, but looking good otherwise :)

src/core/scriptmanager.cpp
dfaure-kdab: I showed Simon how to use r-value references all the way down from start to finish, expect an updated commit ;)

LeonMatthesKDAB: Ah, very nice :)

LeonMatthesKDAB: Note that I will be on vacation starting tomorrow.
@dfaure-kdab Feel free to approve yourself.

You may have to dismiss my old review then, which you should be able to do at the bottom of the PR.
image

src/core/scriptdialogitem.cpp
```diff
@@ -145,6 +148,28 @@ ScriptDialogItem::ScriptDialogItem(QWidget *parent)
...
+        if (it.value().is_string()) {
+            value = QString::fromStdString(it.value().get<std::string>());
+        } else if (it.value().is_boolean()) {
+            value = it.value().get<bool>();
+        } else if (it.value().is_number_integer()) {
+            value = it.value().get<int>();
+        } else if (it.value().is_number_float()) {
+            value = it.value().get<double>();
+        } else {
+            value = QString::fromStdString(it.value().dump());
```
dfaure-kdab, line 166: Maybe with an error message (with spdlog) here about the unsupported data type, to facilitate debugging?

src/core/scriptrunner.cpp
```diff
@@ -346,6 +347,7 @@ QVariant ScriptRunner::runQml(const QString &fileName, QQmlEngine *engine)
                 window->show();
             } else if (auto dialog = qobject_cast<ScriptDialogItem *>(topLevel)) {
                 engine->setProperty("scriptWindow", true);
+                dialog->initialize(std::move(data)); // Pass the json data to initialize method
```
dfaure-kdab, line 350: The comment doesn't really add any value ;)

src/core/scriptdialogitem.cpp
```diff
@@ -557,6 +590,11 @@ void ScriptDialogItem::changeValue(const QString &key, const QVariant &value)
             if (comboBox->count())
                 comboBox->setCurrentIndex(0);
         }
+        return;
+    }
+
+    if (!widget) {
+        spdlog::warn("No widget found for the key '{}'.", key.toStdString());
```
dfaure-kdab, line 597:
```
} else {
    spdlog::warn("Unsupported widget type '{}' for the key '{}'.", widget->metaObject()->className(), key.toStdString());
}
```

dfaure-kdab (approved):

PR #142, opened 2024-08-07
fix: flush spdlog messages also when logging to stderr

smnppKDAB:
The existing code was only enabling flush when logs are redirected to files.
Without this, the relative order of the debug output from qWarnings() and from Message.log() wasn't preserved, making debugging difficult.


src/core/knutcore.cpp
```diff
@@ -207,8 +208,6 @@ void KnutCore::initializeMultiSinkLogger()
         Core::Settings::instance()->logFilePath().toStdString(), SIZE_MAX, max_files, rotate_on_open);
     auto logger = spdlog::default_logger();
     logger->sinks().push_back(fileSink);
-    // auto flush when "info" or higher message is logged.
```
dfaure-kdab, line 210: Why did you lose the comment?

dfaure-kdab (approved):

PR #143, opened 2024-08-07
feat: add a find in files API to `Project`

smnppKDAB:
The methode return a list of document + position of the find items.
The method uses ripgrep (rg) for searching, which must be installed and accessible in PATH.
The `pattern` parameter should be a valid regular expression.
Fixes KDAB#21

### DIFF
--- a/docs/API/knut/project.md
+++ b/docs/API/knut/project.md
@@ -22,7 +22,9 @@ import Knut
 |array&lt;string> |**[allFilesWithExtension](#allFilesWithExtension)**(string extension, PathType type = RelativeToRoot)|
 |array&lt;string> |**[allFilesWithExtensions](#allFilesWithExtensions)**(array&lt;string> extensions, PathType type = RelativeToRoot)|
 ||**[closeAll](#closeAll)**()|
+|array&lt;object> |**[findInFiles](#findInFiles)**(const QString &pattern)|
 |[Document](../knut/document.md) |**[get](#get)**(string fileName)|
+|bool |**[isFindInFilesAvailable](#isFindInFilesAvailable)**()|
 |[Document](../knut/document.md) |**[open](#open)**(string fileName)|
 ||**[openPrevious](#openPrevious)**(int index = 1)|
 ||**[saveAllDocuments](#saveAllDocuments)**()|
@@ -75,6 +77,25 @@ Returns all files with an extension from `extensions` in the current project.
 
 Close all documents. If the document has some changes, save the changes.
 
+#### <a name="findInFiles"></a>array&lt;object> **findInFiles**(const QString &pattern)
+
+Search for a regex pattern in all files of the current project using ripgrep.
+Returns a list of results (QVariantMaps) with the document name and position ("file", "line", "column").
+
+Example usage in QML:
+
+```js
+let findResults = Project.findInFiles("foo");
+for (let result of findResults) {
+    Message.log("Filename: " + result.file);
+    Message.log("Line: " + result.line);
+    Message.log("Column" + result.column);
+}
+```
+
+Note: The method uses ripgrep (rg) for searching, which must be installed and accessible in PATH.
+The `pattern` parameter should be a valid regular expression.
+
 #### <a name="get"></a>[Document](../knut/document.md) **get**(string fileName)
 
 Gets the document for the given `fileName`. If the document is not opened yet, open it. If the document
@@ -85,6 +106,10 @@ If the document does not exist, creates a new document (but don't save it yet).
 
 !!! note
     This command does not change the current document.
+
+#### <a name="isFindInFilesAvailable"></a>bool **isFindInFilesAvailable**()
+
+Checks if the ripgrep (rg) command-line tool is available on the system.
 
 #### <a name="open"></a>[Document](../knut/document.md) **open**(string fileName)
 
--- a/src/core/project.cpp
+++ b/src/core/project.cpp
@@ -28,6 +28,8 @@
 #include <QDirIterator>
 #include <QFileInfo>
 #include <QMetaEnum>
+#include <QProcess>
+#include <QStandardPaths>
 #include <algorithm>
 #include <kdalgorithms.h>
 #include <map>
@@ -378,6 +380,97 @@ Document *Project::openPrevious(int index)
     const QString &fileName = m_documents.at(index)->fileName();
 
     LOG_RETURN("document", open(fileName));
+}
+
+/*!
+ * \qmlmethod array<object> Project::findInFiles(const QString &pattern)
+ * Search for a regex pattern in all files of the current project using ripgrep.
+ * Returns a list of results (QVariantMaps) with the document name and position ("file", "line", "column").
+ *
+ * Example usage in QML:
+ *
+ * ```js
+ * let findResults = Project.findInFiles("foo");
+ * for (let result of findResults) {
+ *     Message.log("Filename: " + result.file);
+ *     Message.log("Line: " + result.line);
+ *     Message.log("Column" + result.column);
+ * }
+ * ```
+ *
+ * Note: The method uses ripgrep (rg) for searching, which must be installed and accessible in PATH.
+ * The `pattern` parameter should be a valid regular expression.
+ */
+QVariantList Project::findInFiles(const QString &pattern) const
+{
+    LOG("Project::findInFiles", pattern);
+
+    QVariantList result;
+
+    const QString path = QStandardPaths::findExecutable("rg");
+    if (path.isEmpty()) {
+        spdlog::error("Ripgrep (rg) executable not found. Please ensure that ripgrep is installed and its location is "
+                      "included in the PATH environment variable.");
+        return result;
+    }
+
+    QProcess process;
+
+    const QStringList arguments {"--vimgrep", "-U", "--multiline-dotall", pattern, m_root};
+
+    process.start(path, arguments);
+    if (!process.waitForFinished()) {
+        spdlog::error("The ripgrep process failed: {}", process.errorString());
+        return result;
+    }
+
+    const QString output = process.readAllStandardOutput();
+
+    const QString errorOutput = process.readAllStandardError();
+    if (!errorOutput.isEmpty()) {
+        spdlog::error("Ripgrep error: {}", errorOutput);
+    }
+
+    const auto lines = output.split('\n', Qt::SkipEmptyParts);
+    result.reserve(lines.count() * 3);
+    for (const QString &line : lines) {
+        QString currentLine = line;
+        currentLine.replace('\\', '/');
+        const auto parts = currentLine.split(':');
+
+        QString filePath;
+        int offset = 0;
+        if (parts.size() > 2 && parts[0].length() == 1 && parts[1].startsWith('/')) {
+            filePath = parts[0] + ':' + parts[1];
+            offset = 2;
+        } else {
+            filePath = parts[0];
+            offset = 1;
+        }
+
+        if (parts.size() > offset + 1) {
+            QVariantMap matchResult;
+            matchResult.insert("file", filePath);
+            matchResult.insert("line", parts[offset].toInt());
+            matchResult.insert("column", parts[offset + 1].toInt());
+            result.append(matchResult);
+        }
+    }
+    return result;
+}
+
+/*!
+ * \qmlmethod bool Project::isFindInFilesAvailable()
+ * Checks if the ripgrep (rg) command-line tool is available on the system.
+ */
+bool Project::isFindInFilesAvailable() const
+{
+    QString rgPath = QStandardPaths::findExecutable("rg");
+    if (rgPath.isEmpty()) {
+        return false;
+    } else {
+        return true;
+    }
 }
 
 } // namespace Core
--- a/test_data/tst_project.qml
+++ b/test_data/tst_project.qml
@@ -33,4 +33,34 @@ Script {
         var rcdoc = Project.open("MFC_UpdateGUI.rc")
         compare(rcdoc.type, Document.Rc)
     }
+
+    function test_findInFiles() {
+        if(Project.isFindInFilesAvailable()) {
+        let simplePattern = "CTutorialApp::InitInstance()"
+        let simpleResults = Project.findInFiles(simplePattern)
+
+            compare(simpleResults.length, 2)
+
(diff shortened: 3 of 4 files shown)


docs/API/knut/project.md
```diff
@@ -75,6 +76,14 @@ Returns all files with an extension from `extensions` in the current project.
 
 Close all documents. If the document has some changes, save the changes.
 
+#### <a name="findInFiles"></a>QVariantList **findInFiles**(const QString &pattern)
+
+Search for a regex pattern in all files of the current project using ripgrep.
+Returns a list of results (QVariantMaps) with the document name and position ("file", "line", "column").
+
+Note : The method uses ripgrep (rg) for searching, which must be installed and accessible
```
dfaure-kdab, line 84: No space before ":" in English (unlike French)

docs/API/knut/project.md
```diff
@@ -75,6 +76,14 @@ Returns all files with an extension from `extensions` in the current project.
...
+in the system's PATH. The `pattern` parameter should be a valid regular expression.
```
dfaure-kdab, line 85: Not necessarily "the system's".

src/core/project.cpp
```diff
@@ -380,4 +382,58 @@ Document *Project::openPrevious(int index)
...
 
+/*!
+ * \qmlmethod QVariantList Project::findInFiles(const QString &pattern)
+ * Search for a regex pattern in all files of the current project using ripgrep.
+ * Returns a list of results (QVariantMaps) with the document name and position ("file", "line", "column").
+ *
+ * Note : The method uses ripgrep (rg) for searching, which must be installed and accessible
+ * in the system's PATH. The `pattern` parameter should be a valid regular expression.
+ */
+QVariantList Project::findInFiles(const QString &pattern)
```
dfaure-kdab, line 393: Could the method be marked const?

If the const is missing only because someone else forgot it on allFilesWithExtensions(), that's not a good reason ;-)

src/core/project.cpp
```diff
@@ -380,4 +382,58 @@ Document *Project::openPrevious(int index)
...
+{
+    LOG("Project::findInFiles", pattern);
+
+    QVariantList result;
+
+    QString path = QStandardPaths::findExecutable("rg");
+    if (path.isEmpty()) {
+        spdlog::error("Ripgrep (rg) executable not found. Please ensure that ripgrep is installed and its location is "
+                      "included in your system's PATH environment variable.");
```
dfaure-kdab, line 402: s/your system's/the/

src/core/project.cpp
```diff
@@ -380,4 +382,58 @@ Document *Project::openPrevious(int index)
...
+        return result;
+    }
+
+    QProcess process;
+    QStringList arguments;
+
+    arguments << "--vimgrep"
```
dfaure-kdab, line 409: Modern C++ allows to do this better with

    const QStringList arguments{"--vimgrep", "-u", etc.};

This way the QStringList is constructed with the right size from the start, and immediately initialized.

src/core/project.cpp
```diff
@@ -380,4 +382,58 @@ Document *Project::openPrevious(int index)
...
+              << "-U"
+              << "--multiline-dotall" << pattern << m_root;
+
+    process.start(path, arguments);
+    if (!process.waitForFinished()) {
+        spdlog::error("The ripgrep process failed to complete: {}", process.errorString());
```
dfaure-kdab, line 415: Oh it completed alright, but it failed ;)
I would remove "to complete"

src/core/project.cpp
```diff
@@ -380,4 +382,58 @@ Document *Project::openPrevious(int index)
...
+        return result;
+    }
+
+    QString output = process.readAllStandardOutput();
+
+    QString errorOutput = process.readAllStandardError();
+    if (!errorOutput.isEmpty()) {
+        spdlog::error("Erreur ripgrep: {}", errorOutput);
```
dfaure-kdab, line 423: Red alert! French text! ;-)

src/core/project.cpp
```diff
@@ -380,4 +382,58 @@ Document *Project::openPrevious(int index)
...
+ * Note : The method uses ripgrep (rg) for searching, which must be installed and accessible
+ * in the system's PATH. The `pattern` parameter should be a valid regular expression.
+ */
+QVariantList Project::findInFiles(const QString &pattern)
+{
+    LOG("Project::findInFiles", pattern);
+
+    QVariantList result;
+
+    QString path = QStandardPaths::findExecutable("rg");
```
dfaure-kdab, line 399: const QString path = ..

(same for many other variables below). Always declare const the variables that aren't modified further down. Clearer, and sometimes faster (no detaching when iterating). And sometimes it catches bugs (typos) ;)

src/core/project.cpp
```diff
@@ -380,4 +382,58 @@ Document *Project::openPrevious(int index)
...
+    if (!errorOutput.isEmpty()) {
+        spdlog::error("Erreur ripgrep: {}", errorOutput);
+    }
+
+    const auto lines = output.split('\n', Qt::SkipEmptyParts);
+    for (const QString &line : lines) {
+        const auto parts = line.split(':');
+        if (parts.size() >= 3) {
+            QVariantMap matchResult;
+            matchResult["file"] = parts[0].split("/").last();
```
dfaure-kdab, line 431: matchResult.insert("file", ...); is better/faster.

This is explained in details in Effective STL, IIRC, I just didn't keep the reference to the item number... Maybe you can find it in the books here ;)

src/core/project.cpp
```diff
@@ -380,4 +382,58 @@ Document *Project::openPrevious(int index)
...
+            matchResult["line"] = parts[1].toInt();
+            matchResult["column"] = parts[2].toInt();
+            result.append(matchResult);
```
dfaure-kdab, line 434: You could do result.reserve(lines.count() * 3) before the loop then, to improve performance.

src/core/project.cpp
dfaure-kdab: Returning only the filename doesn't seem like a good idea to me. What if the caller wants to actually do something with that file? They'll need at least a relative path to the project root to avoid having to search for it recursively. Worse, what if there are multiple files called main.cpp in the project?

Not sure if absolute or relative-to-project-root is better, but for sure just the filename isn't enough.

test_data/tst_project.qml
```diff
@@ -33,4 +33,25 @@ Script {
...
+        let simpleResults = Project.findInFiles(simplePattern)
+
+        compare(simpleResults[0].file, "TutorialApp.cpp")
+        compare(simpleResults[0].line, 21)
+        compare(simpleResults[0].column, 6)
+
+        compare(simpleResults[1].file, "TutorialDlg.h")
+        compare(simpleResults[1].line, 10)
+        compare(simpleResults[1].column, 9)
+
```
dfaure-kdab, line 48: Also check that simpleResults only contains 2 elements? It could misbehave and return false positives after the 2 that you expect.

dfaure-kdab (changes requested):

docs/API/knut/project.md
```diff
@@ -22,6 +22,7 @@ import Knut
 |array&lt;string> |**[allFilesWithExtension](#allFilesWithExtension)**(string extension, PathType type = RelativeToRoot)|
 |array&lt;string> |**[allFilesWithExtensions](#allFilesWithExtensions)**(array&lt;string> extensions, PathType type = RelativeToRoot)|
 ||**[closeAll](#closeAll)**()|
+|QVariantList |**[findInFiles](#findInFiles)**(const QString &pattern)|
```
dfaure-kdab, line 25: I'm guessing QVariantList isn't what a QML/JS developer wants to see here, but I'm not sure what it should say instead... That's a question for the other reviewers.

test_data/tst_project.qml
```diff
@@ -33,4 +33,27 @@ Script {
...
+        compare(simpleResults[0].file, "TutorialApp.cpp")
+        compare(simpleResults[0].line, 21)
+        compare(simpleResults[0].column, 6)
+        compare(simpleResults[1].file, "TutorialDlg.h")
+        compare(simpleResults[1].line, 10)
+        compare(simpleResults[1].column, 9)
+
+        let multilinePattern = "m_VSliderBar\\.SetRange\\(0,\\s*100,\\s*TRUE\\);\\s*m_VSliderBar\\.SetPos\\(50\\);";
+        let multilineResults = Project.findInFiles(multilinePattern)
```
dfaure-kdab, line 53: The comment about checking the length of the array applies to this one too, obviously.

test_data/tst_project.qml
```diff
@@ -33,4 +33,29 @@ Script {
...
+
+    function test_findInFiles() {
+        let simplePattern = "CTutorialApp::InitInstance()"
+        let simpleResults = Project.findInFiles(simplePattern)
+
+        compare(simpleResults.length, 2)
+
+        simpleResults.sort((a, b) => a.file.localeCompare(b.file));
+
+        compare(simpleResults[0].file.split("knut/").pop(), "test_data/projects/mfc-tutorial/TutorialApp.cpp")
```
dfaure-kdab, line 45: What if I clone knut in ~/knut/v1/knut/ ?  Won't this pop() return v1/?

I'm confused, actually. Do you get a full path or a relative path in `file`? This should actually be documented in the docs for your method, BTW.

smnppKDAB: @dfaure-kdab result[0].file currently gives the full path.
pop() returns the last string in the list of strings given by the split.

dfaure-kdab: OK but what if I choose to name my knut source directory "THE_BEST_TOOL_EVER" instead of "knut"? ;)

Why not just give the full path as the expected value? Something like

`compare(simpleResults[0].file, Project.root + "/TutorialApp.cpp")`

docs/API/knut/project.md
dfaure-kdab: Thinking further about it, I guess it should be something like

    `array<map<string, any>>`

docs/API/knut/settings.md
```diff
@@ -28,7 +28,7 @@ The settings are stored in a json file, and could be:
 - int
 - double
 - string
-- array&lt;string>
+- array<string>
```
dfaure-kdab, line 31: regression? Any way to find out if this one still renders correctly?

docs/API/knut/project.md
```diff
@@ -22,6 +22,7 @@ import Knut
 |array&lt;string> |**[allFilesWithExtension](#allFilesWithExtension)**(string extension, PathType type = RelativeToRoot)|
 |array&lt;string> |**[allFilesWithExtensions](#allFilesWithExtensions)**(array&lt;string> extensions, PathType type = RelativeToRoot)|
 ||**[closeAll](#closeAll)**()|
+|array&lt;map&lt;string, any>> |**[findInFiles](#findInFiles)**(const QString &pattern)|
```
dfaure-kdab, line 25: Matt is right, it should be array<object>, actually.
The Javascript type Object is exactly the container for properties of any type that a QVariantMap is.

Now that I think about it, we could have done some type discovery in javascript with Message.log() to find out about this.

docs/API/knut/querymatch.md
```diff
@@ -18,8 +18,8 @@ import Knut
 | | Name |
 |-|-|
 |[RangeMark](../knut/rangemark.md) |**[get](#get)**(string name)|
-|vector<[RangeMark](../knut/rangemark.md)> |**[getAll](#getAll)**(string name)|
-|vector<[RangeMark](../knut/rangemark.md)> |**[getAllInRange](#getAllInRange)**(string name, [RangeMark](../knut/rangemark.md) range)|
+|vector&lt;[RangeMark](../knut/rangemark.md)> |**[getAll](#getAll)**(string name)|
```
dfaure-kdab, line 21: (pre-existing, something for future work) I'm pretty sure this "vector" should be "array" in javascript terms.

dfaure-kdab (dismissed):

src/core/project.cpp
```diff
@@ -380,4 +382,71 @@ Document *Project::openPrevious(int index)
...
+
+    const QString errorOutput = process.readAllStandardError();
+    if (!errorOutput.isEmpty()) {
+        spdlog::error("Ripgrep error: {}", errorOutput);
+    }
+
+    const auto lines = output.split('\n', Qt::SkipEmptyParts);
+    result.reserve(lines.count() * 3);
+    for (const QString &line : lines) {
+        const auto parts = line.split(':');
```
narnaud, line 426: This does not work on Windows, as the path starts with `C:...`
Also, for the path, it returns values like:
`C:/dev/knut/mfc-utils/examples/tutorialapp\TutorialDlg.h:17:9:/// the CTutorialApp::InitInstance function.` with a mix of / and \ - would be nice to add some consistency and use / everywhere (just a replaceAll on the string).

src/core/project.cpp
```diff
@@ -380,4 +382,71 @@ Document *Project::openPrevious(int index)
     LOG_RETURN("document", open(fileName));
 }
 
+/*!
+ * \qmlmethod array<object> Project::findInFiles(const QString &pattern)
+ * Search for a regex pattern in all files of the current project using ripgrep.
+ * Returns a list of results (QVariantMaps) with the document name and position ("file", "line", "column").
```
narnaud, line 388: The QVariantMaps is a good idea, but I feel you need to add a small example in the documentation:

    ```js
    let findResults = Project.findInFiles("foo");
    for (let result of findResults) {
         Message.log("Filename: " + result.file);
         Message.log("Line: " + result.line);
         Message.log("Column" + result.column);
    }
    ```

narnaud (dismissed):

src/core/project.cpp
```diff
@@ -380,4 +382,94 @@ Document *Project::openPrevious(int index)
...
+        QString filePath;
+        int offset = 0;
+        if (parts.size() > 2 && parts[0].length() == 1 && parts[1].startsWith('/')) {
+            filePath = parts[0] + ':' + parts[1];
+            offset = 2;
+        } else {
+            filePath = parts[0];
+            offset = 1;
+        }
+        filePath.replace('\\', '/');
```
dfaure-kdab, line 448: The `replace` needs to happen before the check for '/' on line 441.

dfaure-kdab (changes requested):

dfaure-kdab (approved):

Issue #33, opened 2024-06-19
Rc: Support alignment flags for static (ie labels) during UI conversion


narnaud:
See here: https://learn.microsoft.com/en-us/windows/win32/controls/button-styles

BS_CENTER, BS_BOTTOM...

narnaud:
Closing, this task is wrong as alignment is already supported for static... it's for buttons, but there are no alignment for button text.

PR #150, opened 2024-08-16
support for alignment flags in rc during ui conversion

phyBrackets:
Initial Work for #33

### DIFF
--- a/src/rccore/rc_convert.cpp
+++ b/src/rccore/rc_convert.cpp
@@ -251,12 +251,33 @@ static Widget convertLabel(const Data &data, Data::Control &control, bool useIdF
     Widget widget;
     widget.className = "QLabel";
 
-    if (control.styles.removeOne("SS_RIGHT") || control.type == static_cast<int>(Keywords::RTEXT))
-        widget.properties["alignment"] = "Qt::AlignRight";
-    if (control.styles.removeOne("SS_CENTER") || control.styles.removeOne("SS_CENTERIMAGE")
-        || control.type == static_cast<int>(Keywords::CTEXT))
-        widget.properties["alignment"] = "Qt::AlignHCenter";
+    int alignment = Qt::AlignLeft | Qt::AlignVCenter; // Default alignment
 
+    if (control.styles.removeOne("SS_RIGHT") || control.type == static_cast<int>(Keywords::RTEXT))
+        alignment = Qt::AlignRight | Qt::AlignVCenter;
+    else if (control.styles.removeOne("SS_CENTER") || control.styles.removeOne("SS_CENTERIMAGE")
+             || control.type == static_cast<int>(Keywords::CTEXT))
+        alignment = Qt::AlignHCenter | Qt::AlignVCenter;
+
+    bool wordWrap = true; // Default to word wrap
+    if (control.styles.removeOne("SS_LEFTNOWORDWRAP")) {
+        wordWrap = false;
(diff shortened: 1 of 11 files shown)


src/rccore/rc_convert.cpp
```diff
@@ -251,12 +251,33 @@ static Widget convertLabel(const Data &data, Data::Control &control, bool useIdF
...
-        || control.type == static_cast<int>(Keywords::CTEXT))
-        widget.properties["alignment"] = "Qt::AlignHCenter";
+        alignment = Qt::AlignRight | Qt::AlignVCenter;
+    else if (control.styles.removeOne("SS_CENTER") || control.styles.removeOne("SS_CENTERIMAGE")
+             || control.type == static_cast<int>(Keywords::CTEXT))
+        alignment = Qt::AlignHCenter | Qt::AlignVCenter;
+
+    bool wordWrap = true; // Default to word wrap
+    if (control.styles.removeOne("SS_LEFTNOWORDWRAP")) {
+        wordWrap = false;
```
narnaud, line 264: That's a good, I missed the NO in the word. please keep it.

src/rccore/rc_convert.cpp
```diff
@@ -251,12 +251,33 @@ static Widget convertLabel(const Data &data, Data::Control &control, bool useIdF
...
+    }
+    if (control.styles.removeOne("SS_WORDELLIPSIS")) {
+        alignment |= Qt::TextWordWrap;
+    }
```
narnaud, line 268: It's not word wrap, it's ellipsis. Unfortunately there are no ways to do that easily in Qt, so drop it.

src/rccore/rc_convert.cpp
```diff
@@ -251,12 +251,33 @@ static Widget convertLabel(const Data &data, Data::Control &control, bool useIdF
...
+
+    widget.properties["alignment"] = alignment;
```
narnaud, line 270: Don't change the alignment if there are no options.
Also the property is a text, not an alignment.

src/rccore/rc_convert.cpp
```diff
@@ -723,7 +741,13 @@ Widget convertDialog(const Data &data, const Data::Dialog &d, Widget::Conversion
     }
 
     for (const auto &control : std::as_const(dialog.controls)) {
-        widget.children.push_back(convertChildWidget(data, dialog.id, control, flags & Widget::UseIdForPixmap));
+        Widget childWidget = convertChildWidget(data, dialog.id, control, flags & Widget::UseIdForPixmap);
+
+        if (control.className == "STATIC") {
+            childWidget.className = "QLabel";
+        }
```
narnaud, line 748: Why do you need that? It should be done in `convertControl` no?

src/rccore/rc_write.cpp
```diff
@@ -147,7 +147,54 @@ void writeDialogToUi(const Widget &widget, QIODevice *device)
 
     writeWidget(writer, widget);
 
-    device->write(writer.dump().toUtf8());
+    QQmlEngine engine;
```
narnaud, line 150: No need for all of that, alignment is already handled properly...

narnaud (commented):
I think I'm just going to fix the wordwrapping... alignment is already handled for static, I can't remember why I created this task.

narnaud:
Let's close this one, I fixed the WordWrap in #161

Issue #26, opened 2024-06-19
Add a way to display log in a ScriptDialog


narnaud:
Currently, scripts report their output through logging.
However, the log output doesn't render whilst the script is running.

This could be a good first step to indicate how far along a script is and to show the user that it's indeed doing something.

We could imagine adding a `QPlainTextEdit` with a particular name `LogOutput` or something like that. In the `ScriptDialogItem`, if we find such `QPlainTextEdit`, we initalize it in readonly log mode, and redirect the logging output to it.
Hint: there's a small syntaxhilighting in GUI that could be used, see `LogHighlighter`

LeonMatthesKDAB:
How exactly would this solve our issue? We already have a Log panel.

The issue is that the Log panel isn't updated while JS code runs, as that is single-threaded, right?
But any other UI item, like the mentioned QPlainTextEdit, also won't update, so what would be different here?

narnaud:
The logs are displayed when running a script in the GUI, and visible from the Log panel (we should probably remove them from the history though... another task).

The problem this task is trying to fix is when using knut via the command line: you don't see any logs. For example, when doing the rc to ui conversion, there are some warnings displayed when a flag is not handled, but I'm missing them.

So the idea is to have a way to add a `QPlainTextEdit` inside the script ui that displays the log.
Hope it's clearer.

LeonMatthesKDAB:
Ah, okay. That's clearer.

Shouldn't the command line output just print everything that we dump out with spdlog?

Or do you mean if you don't have a terminal open at all but still run it in script mode?

narnaud:
Unfortunately, it doesn't work on Windows, we can't see the command line output... (limitation of Windows, a GUI program does not display anything when run from the CLI).
The nice thing is that it will also work for people running it from another IDE (like Visual Studio, QtC...)

LeonMatthesKDAB:
Okay, then that clarifies it :+1:

PR #157, opened 2024-08-29
feat: add a way to display log in a ScriptDialog

smnppKDAB:
Added a QPlainTextEdit widget to allow real-time log display in the user interface. The actual code did not allow windows users to view logs when launching knut from the command line.

The QPlainTextEdit can be made visible using the qml setDisplayLogs method and taking as parameter the minimum log level that we want to have.

Fixes https://github.com/KDAB/knut/issues/26

### DIFF
--- a/docs/API/knut/scriptdialog.md
+++ b/docs/API/knut/scriptdialog.md
@@ -22,6 +22,7 @@ import Knut
 ||**[firstStep](#firstStep)**(string firstStep)|
 ||**[nextStep](#nextStep)**(string title)|
 ||**[runSteps](#runSteps)**(function generator)|
+||**[setDisplayLogs](#setDisplayLogs)**(const QString &level)|
 ||**[setStepCount](#setStepCount)**(int stepCount)|
 ||**[verify](#verify)**(bool value, string message = {})|
 
@@ -170,6 +171,11 @@ function convert() {
    runSteps(conversionSteps())
 }
 ```
+
+#### <a name="setDisplayLogs"></a>**setDisplayLogs**(const QString &level)
+
+This method enables real-time logging display in the progress dialog by configuring a QPlainTextEdit widget as a log
+output panel. It filters the logs to show only those that are of the specified log level or higher.
 
 #### <a name="setStepCount"></a>**setStepCount**(int stepCount)
 
--- a/src/core/scriptdialogitem.cpp
+++ b/src/core/scriptdialogitem.cpp
@@ -9,6 +9,7 @@
 */
 
 #include "scriptdialogitem.h"
+#include "loghighlighter.h"
 #include "scriptdialogitem_p.h"
 #include "scriptprogressdialog.h"
 #include "scriptrunner.h"
@@ -44,6 +45,7 @@
 #include <QToolButton>
 #include <QUiLoader>
 #include <QVBoxLayout>
+#include <spdlog/sinks/qt_sinks.h>
 
 namespace Core {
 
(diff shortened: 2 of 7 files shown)


src/core/scriptdialogitem.cpp
```diff
@@ -645,6 +657,29 @@ void ScriptDialogItem::changeValue(const QString &key, const QVariant &value)
...
+        const int index = (it != levels.end()) ? std::distance(levels.begin(), it) : 0;
+        spdlog::level::level_enum logLevel = static_cast<spdlog::level::level_enum>(index);
+        logger->set_level(logLevel);
+
+        logger->sinks().push_back(std::make_shared<spdlog::sinks::qt_sink_mt>(logOutput, "appendPlainText"));
+        for (auto &sink : logger->sinks()) {
+            sink->set_level(logLevel);
+        }
+    }
+    return;
```
dfaure-kdab, line 680: This return saves no purpose, it's the end of the method anyway.

src/core/scriptdialogitem.cpp
```diff
@@ -645,6 +657,29 @@ void ScriptDialogItem::changeValue(const QString &key, const QVariant &value)
...
+
+        const std::vector<QString> levels = {"trace", "debug", "info", "warn", "error", "critical"};
+        const auto it = std::find(levels.begin(), levels.end(), level);
+        const int index = (it != levels.end()) ? std::distance(levels.begin(), it) : 0;
+        spdlog::level::level_enum logLevel = static_cast<spdlog::level::level_enum>(index);
+        logger->set_level(logLevel);
+
+        logger->sinks().push_back(std::make_shared<spdlog::sinks::qt_sink_mt>(logOutput, "appendPlainText"));
+        for (auto &sink : logger->sinks()) {
+            sink->set_level(logLevel);
```
dfaure-kdab, line 677: Why on all sinks, and not just on the one created on line 675?

src/core/scriptdialogitem.cpp
```diff
@@ -645,6 +657,29 @@ void ScriptDialogItem::changeValue(const QString &key, const QVariant &value)
...
+
+        new LogHighlighter(logOutput->document());
+
+        auto logger = spdlog::default_logger();
+
+        const std::vector<QString> levels = {"trace", "debug", "info", "warn", "error", "critical"};
+        const auto it = std::find(levels.begin(), levels.end(), level);
+        const int index = (it != levels.end()) ? std::distance(levels.begin(), it) : 0;
+        spdlog::level::level_enum logLevel = static_cast<spdlog::level::level_enum>(index);
+        logger->set_level(logLevel);
```
dfaure-kdab, line 673: Does this also affect the log level for stderr? Do you need a std::min(oldValue, newValue) maybe, if I understand the API right that the logger has a base level and then the sinks do additional filtering?

src/core/scriptprogressdialog.ui
```diff
@@ -48,6 +48,9 @@
      </property>
     </widget>
    </item>
+   <item row="3" column="0">
+    <widget class="QPlainTextEdit" name="logsPanel"/>
```
dfaure-kdab, line 52: In the Knut main-window there are panels. Here it's just a widget, I suggest renaming it to logsWidget.

src/core/scriptprogressdialog.cpp
```diff
@@ -63,3 +64,9 @@ void ScriptProgressDialog::setReadOnly(bool readOnly)
 {
     ui->buttonBox->setEnabled(!readOnly);
 }
+
+QPlainTextEdit *ScriptProgressDialog::setLogsAvailable()
```
dfaure-kdab, line 68: That's not really a setter. I suggest renaming to logsWidget() and doing the show() in the caller.


(An alternative would be to do everything in this class, i.e. having a `setDisplayLogs(const QString &level)` here, and the one in ScriptDialog just forwards to this one. But I'm OK with it either way, there's already a lot of handling of the progress dialog in ScriptDialog anyway.)

src/core/scriptdialogitem.h
```diff
@@ -19,6 +19,7 @@
 
 class ScriptProgressDialog;
 class QTextDocument;
+class QPlainTextEdit;
```
dfaure-kdab, line 22: unused - remember to re-read your own diffs ;)

docs/API/knut/scriptdialog.md
```diff
@@ -171,6 +172,11 @@ function convert() {
 }
 ```
 
+#### <a name="setDisplayLogs"></a>**setDisplayLogs**(const QString &level)
+
+This method enables real-time logging display within the ScriptDialog by configuring a QPlainTextEdit widget as a log output panel.
```
dfaure-kdab, line 177: It's not really "within the ScriptDialog", now that it's in the progress dialog.

dfaure-kdab (changes requested):

dfaure-kdab (approved):

PR #158, opened 2024-08-30
Deploy Knut on Windows as a zip file

### LINKED ISSUE #29: Create an "installer" from CI
It would be nice to be able to produce a zip with knut and all the libs it's using when creating a release. I don't think we need a real installer with auto-updates...

- [ ] Windows zip (using `windeployqt`
- [ ] Linux ?
- [ ] MacOS ?

LeonMatthesKDAB:
This PR is based on #135 .

Thanks to the great work of @LorenDB , providing a zip file turned out to just be a case of removing the MSIX packaging and uploading the deploy folder.

The resulting executable has been checked by @dfaure-kdab . Knut runs, but with the caveat that Windows complains loudly about the unsigned executable from an unkown vendor. It's also unclear whether we actually need the vc_redist.x64.exe that's included in the zip file.

@narnaud is that acceptable to you?

Closes #29


narnaud:
I don't think we need the icons anymore now that we don't create an installer, you can remove them. Maybe merge all commits into one, to avoid adding/removing files.

Another question I'm wondering: is it creating the artifact at each compilation? Not sure it's needed everytime, but if it's easier like this then let's do it.
And can we add it automatically when we create a release to the release in github?

LeonMatthesKDAB:
I removed the icons again. The PR is set to squash & merge, so will be merged as a single commit.

Its indeed creating an artifact at every commit.
This is on purpose, so we always have a packaged artifact that we can send to users for testing fixes to reported issues.

As this is building with the same preset as the CI for testing, it should be able to perfectly reuse the cache. I'm hoping that means the build will take <1 min. :crossed_fingers: 
We could change this if it ends up taking longer.

The "Hooking on release-please" section suggests it would be possible to upload after an artifact: https://www.amarjanica.com/automate-github-releases-with-release-please
However, I'd prefer to add this once we've gone through a few releases with release-please, as I currently can't test this and would like to see the workflow that runs after merging the release-PR first.

narnaud (approved):

PR #168, opened 2024-09-13
refactor: Remove files with unknown copyright

narnaud:
- replace 2048Game with MainWindow
- replace the mfc-tutorial project with mfc-dialog

Both are created using the MSVC default template for MFC projects.


narnaud (commented):
NOte, I made an error, it should have been 2 commits... too messy to untangle, so lets leave with it as it is.

test_data/tst_rcwriter/qrc_expected_noalias_onlyexist.qrc
```diff
@@ -1,39 +1,4 @@
 
 <RCC>
-    <qresource prefix="/">
-        <file alias="IDR_MAINFRAME_0">res/Toolbar_0.png</file>
```
LeonMatthesKDAB, line 4: That this file now no longer contains any resources is not ideal, as it would mean even total failure to pick up resource files would lead to a passing test.

Can we just add some dummy files next to the cryEdit rc file that can be picked up here?

LeonMatthesKDAB (changes requested):
I have to admit it's very difficult to review this kind of change, but I think this is okay as-is, apart from the one qrc test, which now barely tests anything.

Can we add some dummy files to test that the conversion still works?

test_data/tst_rcwriter/qrc_expected_noalias_onlyexist.qrc
narnaud: Fixed, I used the mainFrame.rc rc file, and remove one of the resources.

Sorry for this big bulk of a change, I wanted to do 2 patches and failed, and it's quite interleaved together. Focus only on the test files, not the test data.

LeonMatthesKDAB (approved):

PR #179, opened 2024-10-07
fix: don't add tooltip item if it's empty + fix encoding

Montel:


src/rccore/rc_convert.cpp
```diff
@@ -160,12 +160,15 @@ static Widget convertPushButton(const Data &data, const QString &dialogId, Data:
         const auto &values = it->values.value(control.id);
         if (values.count() == 1) {
             pugi::xml_document document;
-            const pugi::xml_parse_result result = document.load_string(values.constFirst().toLatin1().constData(),
+            const pugi::xml_parse_result result = document.load_string(values.constFirst().toLocal8Bit().constData(),
```
narnaud, line 163: How does this fix the encoding? toLocal8Bit is very strange to me, and I'm not sure if it supports accent and other things.
But I may be wrong, please explain (in any case, it should have a comment in the code).

narnaud:
Let's abandon this PR.

PR #182, opened 2024-10-07
fix: fix warning about SUBLANG_NEUTRAL

Montel:


src/rccore/rc_convert.cpp
```diff
@@ -931,6 +931,8 @@ QString convertLanguageToCode(const QString &name)
                 return "en_US";
             } else if (subLang == "SUBLANG_ENGLISH_ZIMBABWE"_L1) {
                 return "en_ZW";
+            } else if (subLang == "SUBLANG_NEUTRAL"_L1) {
```
narnaud, line 934: you can remove the if and just keep the else here.

Montel: yep but I want to be sure that there is not other sublang
=> with this patch if there is a ENGLISH/other sublang we have a warning

with a "else" we will not see it.

narnaud: The complete list is here if you want to be exhausted: https://learn.microsoft.com/en-us/windows/win32/tablet/using-the-recognizers-collection
I missed some of them indeed.

Montel: I updated it.

narnaud (approved):

narnaud:
Please remove the merge commit (don't know why it's here)

PR #184, opened 2024-10-07
fix: now we use property name="mfc_id" not id directly.

Montel:
We need to extract it otherwise we can't extract translation


narnaud (commented):
That's too specific to MFC and Knut here, it's not a generic property from ui files.

So the way to do that in javascript is `uiwidget.getProperty("mfc_id")`. If needed, we can add some helper in javascript, but I don't want to make a real property in C++.

By the way, missing the documentation - but that's not the point ;)

Montel:
Ok I closed it

PR #185, opened 2024-10-08
fix: fix crash when ui doesn't exist

Montel:


narnaud:
Please remove the merge commit

PR #186, opened 2024-10-08
fix: fix duplicate list of shortcuts

Montel:


narnaud (approved):
I don't have any duplicates, but the code is shorter and easier like this, so let's go with it.

PR #193, opened 2024-10-21
feat: return document name

Montel:


src/core/document.h
```diff
@@ -50,6 +50,8 @@ class Document : public QObject
 
     Type type() const;
 
+    Q_INVOKABLE const QString documentName() const;
```
narnaud, line 53: This should be a property, like fileName, not an invokable method.

src/core/document.h
```diff
@@ -21,6 +21,7 @@ class Document : public QObject
 {
     Q_OBJECT
     Q_PROPERTY(QString fileName READ fileName WRITE setFileName NOTIFY fileNameChanged)
+    Q_PROPERTY(QString documentName READ documentName NOTIFY documentNameChanged)
```
narnaud, line 24: Reuse fileNameChanged signal: no need for a second signal, as the 2 are linked.

src/core/document.cpp
```diff
@@ -89,6 +89,15 @@ Document::Type Document::type() const
     return m_type;
 }
 
+const QString Document::documentName() const
```
narnaud, line 92: Missing the property documentation (see at the top of the C++ file)

Montel:
I don't understand why it create merge commit...
I don't know how to remove it.

PR #195, opened 2024-10-21
feat: add info about saving file name

Montel:


src/core/document.cpp
```diff
@@ -207,6 +207,7 @@ bool Document::saveAs(const QString &fileName)
         const QFileInfo fi(m_fileName);
         m_lastModified = fi.lastModified();
     }
+    spdlog::info("{}: {}", FUNCTION_NAME, m_fileName);
```
narnaud, line 210: There's alread a `LOG(fileName)` that does exactly that... why do you need one more here?

PR #202, opened 2024-10-24
feat: return caption from specific language

Montel:


src/core/rcdocument.cpp
```diff
@@ -486,6 +486,24 @@ QString RcDocument::stringForDialogAndLanguage(const QString &language, const QS
     }
 }
 
+/*!
+ * \qmlmethod string RcDocument::captionDialogForLanguage(string language, string dialogId)
+ * Returns the string with `id` for the given `language`.
+ */
+QString RcDocument::captionDialogForLanguage(const QString &language, const QString &dialogId) const
```
narnaud, line 493: In Qt world, it's called windowTitle.
I'm not a big fan of "caption", so let's go with windowTitle or dialogTitle, if you want to make clear it's coming from a dialog.

PR #203, opened 2024-10-24
fix: fix crash when we create a new rc file (language is empty => assert)

Montel:


narnaud:
What's the use case for that?

Anyway, there are 2 commits in this PR, completely independent, and it would be nice to have a unit tests if it's fixing a crash.

Montel:
The crash arrived when we use "palette"
for example you want to search file foo.rc, palette will show "Foo.rc" => not the same case
=> when we press enter we will not find it => it will create a foo.rc file.

=> when it will create it (an empty file) => in constructor we use "Q_ASSERT(m_rcFile.data.contains(m_language));"

=> but empty file can't have a m_language !

=> it will assert by default.

We can't really create an autotest for it.

narnaud:
Ok, so the problem is not really the rc file, but the fact that the palette is not working as expected.
I can't really test that on Windows, but it would be better to fix the palette than this.

Note: still 2 independant commits in the same PR

Montel:
It depends if you want to be able to create a new file from palette => it works as expected.
If you don't want to create new file, indeed it's a bug in palette.
So for you palette mustn't create new file ?

narnaud:
No, the palette should never create a new file, just open an existing file in the project.

Montel:
fixed in #213

PR #204, opened 2024-10-24
show info rggrip not found

Montel:
- **fix: fix crash when we create a new rc file (language is empty => assert)**
- **feat: show info/disable line when Ripgrep not found**

### DIFF
--- a/src/gui/findinfilespanel.cpp
+++ b/src/gui/findinfilespanel.cpp
@@ -15,6 +15,7 @@ Contact KDAB at <info@kdab.com> for commercial licensing options.
 #include <QHBoxLayout>
 #include <QHeaderView>
 #include <QLineEdit>
+#include <QPainter>
 #include <QToolButton>
 #include <QTreeWidget>
 #include <QTreeWidgetItem>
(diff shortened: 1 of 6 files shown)


src/gui/findinfilespanel.h
```diff
@@ -12,13 +12,28 @@ Contact KDAB at <info@kdab.com> for commercial licensing options.
...
+
+    bool findInFilesAvailable() const;
+    void setFindInFilesAvailable(bool newFindInFilesAvailable);
+
+protected:
+    void paintEvent(QPaintEvent *) override;
+
+private:
+    bool m_findInFilesAvailable = true;
+};
```
narnaud, line 35: Only used in the C++ find, you can move it there, no need to put it in the header file.

src/gui/findinfilespanel.h
```diff
@@ -35,7 +50,7 @@ class FindInFilesPanel : public QWidget
     void setupToolBar();
 
     QWidget *const m_toolBar;
-    QTreeWidget *m_resultsDisplay;
+    FindInFilesTreeWidget *m_resultsDisplay;
```
narnaud, line 53: This could stay a `QTreeWidget` here, if you move the check to the widget constructor

narnaud:
There are 3 commits in this PR...

src/gui/findinfilespanel.cpp
```diff
@@ -45,6 +46,9 @@ FindInFilesPanel::FindInFilesPanel(QWidget *parent)
     connect(m_resultsDisplay, &QTreeWidget::itemActivated, this, [this](QTreeWidgetItem *item, int) {
         openFileAtItem(item);
     });
+    const bool available = Core::Project::instance()->isFindInFilesAvailable();
+    m_searchInput->setEnabled(available);
+    m_resultsDisplay->setFindInFilesAvailable(available);
```
narnaud, line 51: I don't think you need this here. Just put the check int he `FindInFilesTreeWidget` constructor, and leave the input enabled.

PR #213, opened 2024-10-29
fix: don't create new file from palette

Montel:


src/core/project.h
```diff
@@ -58,7 +58,7 @@ class Project : public QObject
 
 public slots:
     Core::Document *get(const QString &fileName);
-    Core::Document *open(const QString &fileName);
+    Core::Document *open(const QString &fileName, bool onlyExistingFile = false);
```
narnaud, line 61: I would prefer to not change this method, as it's an API.

src/gui/palette.cpp
```diff
@@ -481,7 +481,7 @@ void Palette::addFileSelector()
         model->resetFileInfo();
     };
     auto selectFile = [](const QVariant &path) {
-        Core::Project::instance()->open(path.toString());
+        Core::Project::instance()->open(path.toString(), true);
```
narnaud, line 484: What you could do is query the `FileModel` to check if the file exists here, and only call open if it does.

Montel: ok I will do it.

narnaud:
Close, please have a look at #216

PR #216, opened 2024-11-12
fix: Do not create new files from the palette

narnaud:
The palette is just a way to navigate around, it shouldn't be able to
create new files.
Also keep the palette open if the text does not match (like VS Code).

### DIFF
--- a/src/gui/palette.cpp
+++ b/src/gui/palette.cpp
@@ -444,7 +444,12 @@ void Palette::clickItem(const QModelIndex &index)
         const auto text = ui->lineEdit->text().mid(selector.prefix.length()).simplified();
         if (text.isEmpty())
             return;
-        selector.selectionFunc(text);
+        // If there is a source model (some don't have, like the line selector), we need a match with one item from the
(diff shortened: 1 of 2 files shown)


LeonMatthesKDAB (changes requested):
Okay, I have to admit it took me quite a while to figure out how this prevents new file creation.

I would really appreciate some better documentation there.

src/gui/palette.cpp
```diff
@@ -444,7 +444,10 @@ void Palette::clickItem(const QModelIndex &index)
         const auto text = ui->lineEdit->text().mid(selector.prefix.length()).simplified();
         if (text.isEmpty())
             return;
-        selector.selectionFunc(text);
+        if (m_proxyModel->sourceModel() && !ui->treeView->currentIndex().isValid())
```
LeonMatthesKDAB, line 447: This needs better explanation.

As I'm not too familiar with the palette it took me embarrassingly long to figure out that this is here to prevent creation of new files, if they don't exist.
Please add a comment that explains this.

LeonMatthesKDAB (approved):

PR #219, opened 2024-11-17
fix: Fix combobox's width for API executor

### LINKED ISSUE #164: GUI: fix API executor look
The API executor is a small widget that appears under the document (like the find widget), with Ctrl+Shift+R.
You can see some data if you open it with a C++ file.

Two main issues:
1. the menu is not there, it should be somewhere (maybe under File->Show palette...)
2. if you open it on a txt file (or any other file than a C++ file), then on a C++ file, the combobox with the function name is too small. It should adapt to the size of the items in it.

narnaud:
When changing files with the API executor displayed, the combobox didn't
adapt to its content.

Fix #164

### DIFF
--- a/src/gui/apiexecutorwidget.ui
+++ b/src/gui/apiexecutorwidget.ui
@@ -59,6 +59,9 @@
      <property name="whatsThis">
       <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;API:&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
      </property>
+     <property name="sizeAdjustPolicy">
+      <enum>QComboBox::AdjustToContents</enum>
+     </property>
     </widget>
    </item>
    <item>


src/gui/apiexecutorwidget.cpp
```diff
@@ -39,6 +39,7 @@ APIExecutorWidget::APIExecutorWidget(QWidget *parent)
 {
 
     ui->setupUi(this);
+    ui->apiComboBox->setSizeAdjustPolicy(QComboBox::AdjustToContents);
```
LeonMatthesKDAB, line 42: Shouldn't this rather be changed in the apiexecutorwidget.ui file directly, instead of after construction?

LeonMatthesKDAB (changes requested):
This is definitely improving the UX, but this should be set in the .ui file, not the .cpp file, right?

src/gui/apiexecutorwidget.cpp
LeonMatthesKDAB: Hm, even weirder: for me it's not adjusting the size correctly, even with this patch included???
image

narnaud: Could be in ui or code, no big deal for me. But I moved it to the ui file.
As for your screenshot, unfortunately I think it's a bug in the style. If you tried to:
- open a txt file
- open the API executor line
- open a cpp file
You'll see the combobox change size, it's just not enough to display everything: IMO a style bug.

LeonMatthesKDAB (approved):

PR #220, opened 2024-11-18
feat: Add search for rc files

### LINKED ISSUE #25: Improve search in document
In the GUI, create an API for generic search. Right now we can only search (Ctrl+F) in TextDocument.
It would be good to have the same search line showing up on the different document, with an API to integrate depending on the view:

- [x] Search in QtUiDocument
- [x] Search in QtTsDocument
- [x] Search in the left tree of RcDocument

narnaud:
For now, only search inside the list of data IDs.
We may extend this later on if useful/needed.

To do that, I had to move the rcui code inside the gui library, so I can use the FindInterface.

Fix #25


src/gui/searchableviewinterface.cpp
```diff
@@ -8,40 +8,43 @@
...
-#include "searchabletableview.h"
+#include "searchableviewinterface.h"
 #include "core/textdocument.h"
 #include "highlightdelegate.h"
 
 namespace Gui {
 
-SearchableTableView::SearchableTableView(QWidget *parent)
-    : QTableView(parent)
+SearchableViewInterface::SearchableViewInterface(QAbstractItemView *view)
```
LeonMatthesKDAB, line 17: This name is confusing, as it's not actually an "Interface".
An interface I would expect to inherit from and override methods on (like you do with `FindInterface`).
This doesn't do that, it's rather an adapter/proxy, that allows to conveniently forward the implementation of the `FindInterface` to a `QAbstractItemView`.

Also: "Search" is a new term here, why not just "find"?
As that's used everywhere else.

So maybe `FindAdapter` or `FindProxy`?
The latter would somewhat fit  with the QSortFilter**Proxy**Model

src/gui/rcui/rcfileview.cpp
```diff
@@ -9,6 +9,7 @@
 */
 
 #include "rcfileview.h"
+#include "../searchableviewinterface.h"
```
LeonMatthesKDAB, line 12: prefer `"gui/searchableviewinterface.h"`

LeonMatthesKDAB (changes requested):
Overall the change is okay, but the `SearchableViewInterface` name is confusing, as this class implements "Find", not "Search" and actually isn't an "interface".

src/gui/searchableviewinterface.cpp
narnaud: I agree, it was an interface at first, and I changed mid-way. `FindAdapter` is way better.

LeonMatthesKDAB (approved):

PR #222, opened 2024-11-19
fix: Fix json loading in gui

narnaud:
The loading was working to be able to use the JsonDocument API, but it
didn't work in the GUI to display the file.

This is fixed, and we update the underlying json model after each save.


LeonMatthesKDAB (approved):
Overall this fixes the problem that JSON documents aren't loading at all.

However, I want to point out that this is still not bug-free!
If I see it correctly, editing the document in Knuts text view won't update the m_jsonData.
These changes therefore won't be reflected in the `::value`/`::hasValue` calls.
`::setValue` will also override any changes to the document that weren't saved.

Overall, the `JsonDocument` still needs some more work, but this is moving in the right direction.

narnaud:
I agree with you, there is still an issue syncing both representation of the json file.
I fixed the most pressing issue, but for now you can't mix'n match `JsonDocument` API and `TextDocument` API.
I've create #226 .

PR #223, opened 2024-11-20
refactor: Chage back the run dialog wiodget to a dialog

narnaud:
The widget is not needed anymore now that we have interactive scripts.
Note: need update of the photon scripts (later).

Revert "Move RunScript widget as part of the main window"
This reverts commit b2793abbedd25a1f26b1e46dfbbb089e2aa7614a.

New features:
- use a combobox to list the scripts (easier to find)
- save/restore latest run script (useful particularly with IDE integration)


LeonMatthesKDAB (commented):
From A UI/UX perspective I'm in favor of this change.
The dialog grabs your attention a lot better than the "Run Widget", which I missed popping up a couple of times and was confused that nothing happened.

However, I don't quite understand the motivation for making this a widget in the first place. Especially what this has to do with interactive scripts?
Are we missing something here?
@narnaud Can you elaborate what this has to do with interactive scripts?

narnaud:
Yes, originally the run script was a dialog like that.
It has been changed later on when we integrated the Photon script, as the script is basically run all the time. So having a small line like the find widget with a button you can re-click made sense.

For photon, we could use interactive scripts to run the script another time, or stop it. Or we could simply have a UI script with a button "Run". We don't have to change the run script dialog for that.

LeonMatthesKDAB (approved):
@narnaud Okay, I get the point now.
Scripts can build their own UI, so if they need to run multiple subsequent steps, that's the better route to go.

PR #236, opened 2025-01-22
chore: Remove unused include

Montel:


iamsergio:
Note that every "feat: " commit msg will appear in the release changelog
For an include you'll want "chore: "

PR #237, opened 2025-01-22
feat: Add basic support for Rust documents

Montel:
Currently, this only includes basic support for the TreeSitter grammar,
without support for symbols or LSP.


narnaud (changes requested):
Looks good, but like the previous one please remove the merge commit.

PR #239, opened 2025-02-06
Fix memory leaks

Cuperino:
Fix memory leaks regarding:
- Code file viewer
- UI file viewer
- Slint file viewer
- Image file viewer
- TS translation file viewer
- Script panel

### DIFF
--- a/src/gui/qmlview.cpp
+++ b/src/gui/qmlview.cpp
@@ -31,7 +31,8 @@ QmlView::QmlView(QWidget *parent)
     addAction(action);
 }
 
-QmlView::~QmlView() {
+QmlView::~QmlView()
+{
     for (auto qmlView : std::as_const(m_qmlViews)) {
         qmlView->close();
         delete qmlView;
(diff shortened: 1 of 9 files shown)


src/gui/qmlview.cpp
```diff
@@ -41,7 +41,7 @@ void QmlView::runQml()
 
     const QString qmlFilePath = document()->fileName();
 
-    auto *qmlView = new QQuickView();
+    auto *qmlView = new QQuickView(this->window()->windowHandle());
```
narnaud, line 44: It doesn't work as expected.
It should show a new window (I suppose that's the reason why there wasn't any parents).
A way to fix that would be to use Delete_on_Close widget attribute.

narnaud (changes requested):
Nice catch for most of them, except the QML preview window ;)

narnaud:
Btw, make sure to use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for your commit, and properly initialize pre-commit:
```
pre-commit install --hook-type commit-msg
```

Cuperino:
> Btw, make sure to use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for your commit, and properly initialize pre-commit:
> 
> ```
> pre-commit install --hook-type commit-msg
> ```

Thanks!

narnaud (approved):

PR #240, opened 2025-02-06
Make document tabs closable

Cuperino:
Allow closing documents by clicking X in their tabs.

### DIFF
--- a/src/gui/knuttabwidget.cpp
+++ b/src/gui/knuttabwidget.cpp
@@ -18,27 +18,18 @@ namespace Gui {
 
 KnutTabBar::KnutTabBar(QWidget *parent)
     : QTabBar(parent)
-    , m_hideCloseTimer(new QTimer(this))
 {
-    // Human moment delay before hiding close buttons
-    m_hideCloseTimer->setInterval(2000);
-    connect(m_hideCloseTimer, &QTimer::timeout, this, [this]() {
-        setTabsClosable(false);
-    });
 }
 
 bool KnutTabBar::event(QEvent *event)
 {
     switch (event->type()) {
-    case QEvent::HoverEnter:
-        m_hideCloseTimer->stop();
-        setTabsClosable(true);
-        break;
-    case QEvent::HoverLeave:
-        m_hideCloseTimer->start();
-        break;
     case QEvent::Show:
+    case QEvent::HoverLeave:
         setTabsClosable(false);
+        break;
+    case QEvent::HoverEnter:
+        setTabsClosable(true);
         break;
     default:
         break;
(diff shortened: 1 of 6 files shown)


Cuperino:
> I didn't do that because I find it very ugly...

This sounds more like a theming and Qt design issue... This what the buttons look like in KDE Plasma's Breeze dark theme:
image
image
I don't like that it's inconsistent with the close button from the sidebar, but it doesn't look too bad on this instance.

> - only show the button on the current tab
> - or when hovering the other tabs
> - change the icon to use the close icon from the dock for example

I tried going with your first option, for I think that is the best possible solution. However, tabList is stored in the
protected side of QTabBar (qtabbar_p.h), and there's no good way to access it without re-inventing the wheel. I ended up implementing your second suggestion instead: [show] when hovering the other tabs.

Having all tabs appear and disappear at the simultaneously can be jarring if it happens consecutively too quickly, so I also added a timer before the buttons are hidden again. For the delay I chose 2 seconds, which is what is used for a "human moment" over at KDE.

It's important for users to have an intuitive and efficient way to close tabs with the mouse cursor. I also thought about closing a tab with a middle click, but that wouldn't have been as intuitive as having a close button in the tab.

src/gui/knuttabwidget.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include <QTimer>
+
+namespace Gui {
+
+KnutTabBar::KnutTabBar(QWidget *parent)
+    : QTabBar(parent)
+    , m_hideCloseTimer(new QTimer(this))
+{
+    // Human moment delay before hiding close buttons
+    m_hideCloseTimer->setInterval(2000);
```
narnaud, line 24: Testing it, I think I would prefer without the timer.
Interestingly, it works with the default theme, but not with fusion or Oreo on Windows. I'm fine with that, as it means I don't see them ;)

narnaud (changes requested):
Thank you, just rebase and remove the timer, and it's good to go.

src/gui/knuttabwidget.cpp
Cuperino: Interesting. I get the same experience in Plasma and in MacOS. It think it has to do with the themes being, possibly, incomplete.

On Plasma, when I switch to either of those theme from the default theme, the feature works but it displays Windows7-like, orange, close buttons that feel out of place. When I load the app from the start with either of those themes, the feature simply doesn't work.

Over on MacOS, the feature only ever works on the default theme. The MacOS default theme also shows the close button on one tab at a time, which is a nice quality other systems don't have, but then the numbers to the left of the text editor look out of place. The Fusion theme feels right at home in MacOS.

Regardless of any of that, I've removed the timer. Should be ready to merge.

narnaud (approved):

PR #250, opened 2025-03-20
feat: Auto indent when Enter is pressed in ScriptPanel

Cuperino:
This change is meant to boost productivity when prototyping a Knut script from the ScriptPanel.

Behavior aims to replicate what one finds in Qt Creator and standard editors. Pressing Enter|Return results in a line that has been properly indented relative to the current line. When a new block is about to begin, a new indentation is added. Blocks are determined by checking for Javascript verbs that result in a block and by checking for an opening curly brace.

When an opening brace is typed, the closing brace that follows is inserted automatically. Typing a closing brace after a closing brace was automatically inserted results in the cursor being moved to the right of the automatically typed brace. Pressing Enter after a right brace was automatically inserted results in the closing brace being properly re-indented bellow the new line.

I've replaced use of if/else statements with a switch, so the compiler can do a better job at optimizing this section.

### DIFF
--- a/src/gui/scriptpanel.cpp
+++ b/src/gui/scriptpanel.cpp
@@ -336,13 +336,100 @@ void ScriptPanel::mouseReleaseEvent(QMouseEvent *mouseEvent)
 
 void ScriptPanel::keyPressEvent(QKeyEvent *event)
 {
-    if ((event->key() == Qt::Key_Backtab)) {
+    switch (event->key()) {
+    case Qt::Key_Backtab:
         Core::indentTextInTextEdit(this, -1);
         return;
-    } else if (event->key() == Qt::Key_Tab) {
+    case Qt::Key_Tab:
         Core::indentTextInTextEdit(this, 1);
         return;
-    } else if (event->key() == Qt::Key_F1) {
+    case Qt::Key_BracketLeft: // '['
+    case Qt::Key_BraceLeft: { // '{'
+        // Auto add right brace
+        QPlainTextEdit::keyPressEvent(event);
+        QTextCursor cursor = textCursor();
+        const QString key = (static_cast<QChar>(event->key() + 2));
+        cursor.insertText(key);
+        cursor.movePosition(QTextCursor::PreviousCharacter);
+        setTextCursor(cursor);
+        return;
+    }
+    case Qt::Key_BracketRight: // ']'
+    case Qt::Key_BraceRight: { // '}'
+        // Prevent right brace from being added twice if it's already been added
+        QTextCursor cursor = textCursor();
+        cursor.movePosition(QTextCursor::NextCharacter, QTextCursor::KeepAnchor);
+        const QString key = static_cast<QChar>(event->key());
+        if (cursor.selectedText() == key) {
+            cursor.clearSelection();
+            setTextCursor(cursor);
+            return;
+        }
+        break; // Exit switch and handle press
+    }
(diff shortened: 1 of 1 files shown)


src/gui/scriptpanel.cpp
```diff
@@ -336,13 +336,93 @@ void ScriptPanel::mouseReleaseEvent(QMouseEvent *mouseEvent)
...
+        cursor.movePosition(QTextCursor::PreviousCharacter);
+        setTextCursor(cursor);
+        return;
+    }
+    case Qt::Key_BraceRight: {
+        // Prevent right brace from being added twice if it's already been added
+        QTextCursor cursor = textCursor();
+        cursor.movePosition(QTextCursor::NextCharacter, QTextCursor::KeepAnchor);
+        if (cursor.selectedText() == "}") {
+            cursor.setPosition(cursor.anchor());
```
narnaud, line 360: I think you should move it after the `}`, not before.
It's very strange that I type a letter and it does nothing.

src/gui/scriptpanel.cpp
```diff
@@ -336,13 +336,93 @@ void ScriptPanel::mouseReleaseEvent(QMouseEvent *mouseEvent)
...
+                if (!newBlock) {
+                    cursor.movePosition(QTextCursor::EndOfLine);
+                    cursor.movePosition(QTextCursor::PreviousCharacter, QTextCursor::KeepAnchor);
+                    while (cursor.selectedText() == " ") {
+                        cursor.clearSelection();
+                        cursor.movePosition(QTextCursor::PreviousCharacter, QTextCursor::KeepAnchor);
+                    }
+                    if (cursor.selectedText() == "}") {
+                        cursor.clearSelection();
+                        cursor.movePosition(QTextCursor::PreviousCharacter, QTextCursor::KeepAnchor);
```
narnaud, line 405: You need to break here too.
For example, if you type `if () {}` and enter, the line will have too much indentation.

src/gui/scriptpanel.cpp
```diff
@@ -336,13 +336,93 @@ void ScriptPanel::mouseReleaseEvent(QMouseEvent *mouseEvent)
...
+    switch (event->key()) {
+    case Qt::Key_Backtab:
         Core::indentTextInTextEdit(this, -1);
         return;
-    } else if (event->key() == Qt::Key_Tab) {
+    case Qt::Key_Tab:
         Core::indentTextInTextEdit(this, 1);
         return;
-    } else if (event->key() == Qt::Key_F1) {
+    case Qt::Key_BraceLeft: {
```
narnaud, line 346: This is nice, could you do that also for `[]` too?
Using the event, it should be possible to do that easily without any code duplication.

narnaud (changes requested):

src/gui/scriptpanel.cpp
Cuperino: Nice catch! Done.

src/gui/scriptpanel.cpp
Cuperino: Fixed. My mistake. I forgot to update the document's cursor and it wouldn't update as a result. Set position was also redundant there.

src/gui/scriptpanel.cpp
Cuperino: Nice catch. This helped me notice other similar bugs. Fixed it in a different way because of that.

PR #253, opened 2025-03-29
Add Script Panel Shortcuts

Cuperino:
- feat: Add New, Open, and Save shortcuts to Script Panel
- fix: Check whether script needs to be saved before opening a new one

### DIFF
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -357,10 +357,14 @@ void MainWindow::showEvent(QShowEvent *event)
 
 void MainWindow::openProject()
 {
-    auto path = QFileDialog::getExistingDirectory(this, tr("Open project"), QDir::currentPath());
-    if (!path.isEmpty()) {
-        Core::Project::instance()->setRoot(path);
-        initProject(path);
+    if (m_scriptPanel->hasFocus())
+        m_scriptPanel->openScript();
+    else {
+        auto path = QFileDialog::getExistingDirectory(this, tr("Open project"), QDir::currentPath());
+        if (!path.isEmpty()) {
+            Core::Project::instance()->setRoot(path);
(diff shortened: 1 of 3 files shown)


src/gui/mainwindow.cpp
```diff
@@ -510,9 +514,13 @@ void MainWindow::reloadDocuments()
...
-    auto document = Core::Project::instance()->currentDocument();
-    if (document)
-        document->save();
+    if (m_scriptPanel->hasFocus())
+        m_scriptPanel->saveScript();
+    else {
+        auto document = Core::Project::instance()->currentDocument();
+        if (document)
+            document->save();
+    }
```
narnaud, line 523: I would prefer to not change the mainwindow.cpp - after all, those actions are specific to the script panel.

We could do that with a `QShortcut` in `ScriptPanel`:
```
    auto shortcut = new QShortcut(QKeySequence::Open, this);
    shortcut->setContext(Qt::WidgetWithChildrenShortcut);
    connect(shortcut, &QShortcut::activatedAmbiguously, this, &ScriptPanel::openScript);
```
Or, maybe better, create actions on the tool buttons and assign non-ambiguous shortcuts (Shift+Ctrl+...).

src/gui/scriptpanel.cpp
```diff
@@ -429,6 +432,12 @@ void ScriptPanel::keyPressEvent(QKeyEvent *event)
         }
         return;
     }
+    case Qt::Key_N:
```
narnaud, line 435: This can be done in one line in the constrtuctor with:
```
    newScriptAction->setShortcut(QKeySequence::New);
```
If you are afraid of ambiguous shortcuts, see for the others

narnaud (changes requested):

PR #257, opened 2025-09-10
fix: cmake 4.0 support doesn't support cmake < 3.5 by default

Montel:
nlohmann-json still requires cmake 3.1
it's not supported by cmake 4.0

We need to use CMAKE_POLICY_VERSION_MINIMUM=3.5


narnaud:
If it's about nlohmann-json, we should update to the last version, 3.12.0, which adds support to CMake 4.

